PR #110, opened 2016-08-12
Fixed setup script

JanWerder:
Hey Mike,

I tried to start up Dairybox on my Windows 10 machine and that didn't work. The setup.sh turned out to be the problem. Getting the path in the fashion the old script did it, didn't work, so I replaced them with fixed paths. So now it works, but when I try to display http://192.168.48.48:8084/ I do get the following:

image

Not sure what causes that message to be there for over 10 minutes now, but that would be a different problem, I guess.

### DIFF
--- a/scripts/apache/add-ports.sh
+++ b/scripts/apache/add-ports.sh
@@ -2,9 +2,7 @@
 # This script adds the extra ports for other Ludum Dare services.
 # i.e. 8080 for api.ludumdare.com, 8081 for static.ludumdare.com, etc.
 
-SCRIPT=$(readlink -f "$0")
-SCRIPTPATH=$(dirname "$SCRIPT")
-cd $SCRIPTPATH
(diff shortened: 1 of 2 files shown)


povrazor:
My apologies. Mainline is broken right now, and may not work on Windows for a while. I'm right in the middle of reworking the toolchain, hence all the commits labeled "broken".

The "Update in progress" message means the code hasn't been compiled yet. If you were on Linux, you might be able to nudge it and make it work like so, but I haven't tested it on another machine yet (literally just got it working last night).

```
npm install
make
```

Eventually that'll be hidden from you and done automatically by the VM, but I haven't got that far yet. Still need to finish the SVG pipeline.

JanWerder:
Okay, that's alright. I tried that but it didn't work out. I'll just wait until the toolchain is fixed. No hurry!

PR #294, opened 2016-11-28
Introduce MIT license

leereilly:
This PR introduces the [MIT license](http://choosealicense.com/licenses/mit/), but IANAL and this isn't my project ¯\\_(ツ)_/¯ 

/cc @povrazor


povrazor:
Sure. That was the plan anyway.

Issue #325, opened 2016-11-30
Theme Slaughter: Add editing


mikekasprzak:
Like it used to be, it would be nice if you could click on a theme in your Previous Themes list, and change your vote.

noah-potter:
Would this work like a toggle? i.e. clicking on the 'x' symbol would immediately change the vote to YES and show the checkmark?

povrazor:
No, it would change what you're currently voting on (as it used to do).

noah-potter:
Gotcha. I haven't used the old site very much so I'll be asking basic questions about how things worked, sorry. I'd like to help but I can't find much setup documentation. I'll keep an eye on the repo for that though.

povrazor:
Yeah, setup is extra broken at the moment. It should be in a better state in the new year.

gonutz:
I want to support this issue. Being unable to change my mind after I said no to a theme has bugged me the last two jams. I have even stopped slaughtering themes this time because the way I do it - probably the same as everybody else - is that I look at a theme, then think of the first idea that comes to my mind and make my decision based on that. But then two themes later I still think about the previous theme and a cool idea pops up in my head. But by then I have already rejected the theme.
For my use-case it would be enough to be able to edit the last two or three themes. Right now I can see the last 10 themes that I slaughtered so being able to just click on the checkmark/cross to toggle the decision would be great.

pfefferminzz:
It's pretty annoying that you can't change your mind even for the recent themes. A toggle would be very useful.

I even flagged a theme as inappropiate by accident and had no possibility to correct myself.

Issue #332, opened 2016-11-30
Theme Slaughter: Previous Theme order is messed up after refresh (was List of past votes stopped updating)


CyberStarLight:
At around 550 votes my theme list stopped adding the new votes. I investigated a bit in the firefox inspector, the vote goes through and I get a response of code 200(OK), and the count of the total votes go up.
but the list of text themes stays the same. when refreshing the page, or even after logging in again.

I'm not sure if my votes are added properly to the database or not.

I looked at the request/response for the list, and it attempts to send all of the 500+ votes names and yes/no. but the new votes are not there. maybe it's trying to send too much?
I guess you could just send the last ~10 votes or so, as all of them don't show anyway. in case it's just a display issue.

povrazor:
Can I get you to add a couple more votes?

I'm seeing 572 suggestions in the database under your ID (761).

CyberStarLight:
You seem to be right, and it's just an ordering issue. I voted again for "Mom's spaghetti", the visible list was not updated but I found it in the response text (after a refresh): as "3626":"Mom's spaghetti". so it does get to the browser, just not displayed.

povrazor:
Cool. So it's the screwy way I fetch the list of previous themes that's to blame.

Issue #301, opened 2016-11-28
Toolchain: Replace MySQL with MariaDB


povrazor:
Okay. Herein the toolchain is extra broken. Setup scripts need to be adjusted to to get us a newer version of MySQL. ScotchBox ships with `MySQL 5.5`, but we should be using `MariaDB 10.1`.

Remove MySQL like so:

```
vagrant ssh
sudo apt-get remove mysql-server
```

Then follow the instructions here to install MariaDB.

https://downloads.mariadb.org/mariadb/repositories/#mirror=digitalocean-nyc

**DO NOT** change the passwords (just push enter when prompted). NOTE: automating this part is what makes adding this to the setup scripts tricky. :)

You can check the current version of MySQL/MariaDB by doing

```
mysql --version
```

And if you don't follow database things, MariaDB is a drop-in replacement for MySQL, so all the commands are exactly the same as before.

Issue #221, opened 2016-10-09
Toolchain: Issues following a dev environment setup


mkalam-alami:
Hi,

I've seen the big "BROKEN" warning in the readme but you may be interested in this list of little tweaks I had to do to get things to run (after setting things up as documented):
### Outside the VM
- `cd www/src/tools && chmod 777 *`
- (if the VM was up) `vagrant halt && vagrant up`
### Inside the VM

```
sudo npm install -g svgo less clean-css buble rollup uglifyjs
cd /var/www
npm install
sudo chmod 777 src/tools/*
make
sudo service apache2 restart # I also renamed the Apache configs to remove an extra ^M, not sure if it was needed 
```

I didn't look into it too much but the `all.min.svg` files didn't appear in my `-` folder, though the `svgo` commands seemed to run without errors. So unfortunately I couldn't enjoy the new LD logo in its full SVG glory :P

povrazor:
Ah yes, I have been saying it's broken, but it's not actually broken, just not documented.

Using the `~/vvv` folder should have proper permissions.

There's a few tools that needed newer versions, from 3rd party repos, I just can't remember which ones. SVGO might have been completely broken (that seems to ring a bell). Either that, or it needed an actual current version of NPM and Node (not what ships in the Ubuntu repos). That said, I might just be remembering the hell that was trying to get Grub to support multiple build targets, hence why it's now a Makefile.

On my setup, I have them installed *outside* the VM, and do my builds there. There was a reason for that, but I can't remember what. I run Ubuntu 16.04.

povrazor:
Some notes on installing the toolchain from outside the VM.

(I'm not actually sure which is preferred, inside or out, but I have to get one of my PC's working again)

* Install Dairybox, and follow the instructions there for setting up your repo
* Make sure `node` and `npm` are installed. Depending on the node version, sudo may not be needed.
* From `www`, install tools
  * `npm install` (this installs 
  * `sudo npm install -g svgo`
  * `sudo apt-get install php`
  * `sudo npm install -g less clean-css buble rollup`

... still working on this. I hit a brick wall, the version of NODE ubuntu ships with is too old.

povrazor:
Alright, the above added instructions should work for anyone attempting to install the toolchain. My bad, but it looks like I forgot to push an update to the toolchain itself from way back in August. It was in my repo, but not mainline Ludum Dare's.

povrazor:
[...]

PR #338, opened 2016-12-02
Allows vote to be changed by clicking recent vote

StormBurpee:
Proposes a fix for issue #325

### DIFF
--- a/src/com/content-event-slaughter/event-slaughter.js
+++ b/src/com/content-event-slaughter/event-slaughter.js
@@ -25,8 +25,10 @@ export default class ContentEventSlaughter extends Component {
 		this.submitFlagVote = this.submitFlagVote.bind(this);
 
 		this.openLink = this.openLink.bind(this);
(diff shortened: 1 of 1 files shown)


povrazor:
Is this tested on the VM (sorry, I haven't finished fixing the setup instructions, or explained how to create an event/set its mode to voting)? It's missing a bind in the constructor, so `this` isn't going to be set correctly in the functions.

Off hand, a `recent.pop()` wont likely work either, as that would remove the last element added, and not the specific one clicked.

Plus #332 probably messes with this.

StormBurpee:
I'll have to try and tinker with it again and get the VM setup properly. I've been just debugging on the website cloning, and using the console. I'll get back to you on that.

I noticed that with the pop, after I submitted the request - I'm sorry about that.
Currently working on fixing this - sorry.

povrazor:
Either way, I appreciate you trying (despite the broken setup).

You can take a look at #221 and #301 for notes a couple things I haven't fixed/documented yet.

Issue #314, opened 2016-11-29
"Invalid Date" when hovering over calendar


pta2002:
![2016-11-29_22-32-12](https://cloud.githubusercontent.com/assets/7443916/20731783/be201486-b683-11e6-9447-4e19abf9734e.png)

povrazor:
What browser?

joaolucasbig:
I'm running on Chrome. Dont get this error here.

pta2002:
It's Firefox. Might see if I can fix it when I get home

povrazor:
Confirmed. I also see it in Firefox.

StormBurpee:
Rather than displaying Invalid Date, the calendar appears to be a month ahead for me - it's showing January dates, rather than December.

joaolucasbig:
Confirmed the @StormBurpee comment: The calendar is in fact a month ahead.
![screenshot_2](https://cloud.githubusercontent.com/assets/5459797/20835293/e64dd0b6-b880-11e6-9e7d-2fae7f56bd45.jpg)

pta2002:
Huh. They are probably related.

StormBurpee:
I think it's to do with the fact that Months are represented from 0-11 in JavaScript. In the calendar.js file (Under src/components/sidebar-calendar)
At the begging of the file PoV mentions this, and adds 1 to make it easier to read through the development. 

> let thisMonth = today.getMonth()+1;
let thisYear = today.getFullYear();
		

> /* Months and Weeks start at 0. Years and Days start at 1. Using 0th day is like -1 */
> let monthEndsOn = new Date(today.getFullYear(), today.getMonth()+1, 0).getDate();
> let nextDay = today.getDate() - today.getDay();

However, later on when the calendar is rendering, it calls the Generate Rows function,
Where he creates another Date variable off of this, (where it's still +1)

> props.title = col.month+"-"+col.day+"-"+col.year;
> if ( window.Intl ) {
> 	// http://stackoverflow.com/a/18648314/5678759
> 	let objDate = new Date(props.title);
> 	props.title = objDate.toLocaleString("en-us", {month:"long", day:"numeric", year:"numeric"});
> }

in the props.title if you subtract 1 from col.month, this should fix the issue. As javascript will parse 12 as 0, as it is 1 over the maximum (11). Therefore, 0 == January

pta2002:
Alright then, that should also fix the Firefox issue (since 12 wouldn't be
a valid month)

[...]

PR #339, opened 2016-12-02
Fixed the Calendar Issue #314

StormBurpee:
this fixed issue #314
Removed the addition to the months, as it was not being converted correctly. It now displays the calendar 100% correctly - (even with the trophies, that's a neat feature PoV!)

image


povrazor:
Yes, this fixes Chrome. Duh, somehow I didn't even notice this was broken.

PR #340, opened 2016-12-02
Fix for issue #314 in Firefox and other browsers

StormBurpee:
**Fix Issue #314**

In firefox, and other earlier browsers it seems as though they haven't fully implemented the map functionality. Through setting the index of the Col, this should allow it to access the correct variables inside the object.
Whilst also experimenting, I found that setting the Date from a string wasn't working and was only supported in Chrome. As you can see, I've manually set it up to put the year, month and day in, rather than doing it from the string.

### DIFF
--- a/src/com/sidebar-calendar/calendar.js
+++ b/src/com/sidebar-calendar/calendar.js
@@ -54,7 +54,9 @@ export default class SidebarCalendar extends Component {
 	}
 
 	genRow( row ) {
-		return row.map( function(col) {
+		return row.map( function(col, index) {
+			if(col[index])
+				col = col[index];
 			let props = {};
 			if ( col.selected ) {
 				props.class = "selected";
@@ -66,10 +68,10 @@ export default class SidebarCalendar extends Component {
 			props.title = col.month+"-"+col.day+"-"+col.year;
 			if ( window.Intl ) {
 				// http://stackoverflow.com/a/18648314/5678759
-				let objDate = new Date(props.title);
+				let objDate = new Date( col.year, col.month-1, col.day );
 				props.title = objDate.toLocaleString("en-us", {month:"long", day:"numeric", year:"numeric"});
 			}
-			
+
 			var ShowIcon = null;
 			if ( col.year == 2016 && col.month == 12 && (col.day >= 9 && col.day <= 12) ) {
 				ShowIcon = <SVGIcon class="-icon">trophy</SVGIcon>;


povrazor:
Do you have a reference on Firefox's missing Map features? I'm curious if a Polyfill would help.

StormBurpee:
Sorry man, It was 3 in the morning here when I wrote that and mustn't have been thinking clearly. I think a Polyfill should work. What I think I was trying to say was that Chrome was able to deduce the variable itself without declaring the index, whereas all the other browsers needed the index to be defined.

povrazor:
Oh I see, so I'm just using map incorrectly.

StormBurpee:
Yeah, in a way... But it's not an obvious mistake to make. I couldn't understand why it wasn't working for so long; the give away was that if you logged the dates they would be undefined.

povrazor:
```
[10,11,12].map(e => 'number: '+e);
```

Firefox doesn't seem to have issues for numbers. Not sure about date objects.

povrazor:
Okay, I mistakenly thought that was a date object, but I don't think it is.

```
[new Date(), new Date()].map(d => d.getMonth());
```

That seems to work fine. As do Objects.

```
[{id:10,text:'hey'},{id:20,text:'ho'}].map(d => d.id+' '+d.text);
```

StormBurpee:
So what's happening with the website is, you're parsing the data array
Which is actually
[[[day: 27, month: 11, year: 2016], [day: 27, month: 11, year: 2016], etc], [], []]
So in your map you have, Parent Array, with three children, those three children have 7 children, and then those children have an object array.
So your array is three dimensions, and in Firefox the map is only parsing the first two dimensions of the array, ignoring the object completely; which is where you have your four variables stored, day, month, selected and year..

StormBurpee:
I don't know I'm not making much sense ahha, it's getting close to 5:30 am!

povrazor:
Yeah that makes sense. It sounds like the caller is making a mistake, or we found an [ICE](https://en.wikipedia.org/wiki/Compilation_error#Internal_Compiler_Errors). No worries. It's a minor bug. Get some sleep. :)

StormBurpee:
Yeah! Could be an ICE, haha I'll get some sleep when I get the VM up and running ;)

povrazor:
Yeah, most of the ICE's I've found have been related to the Buble's JSX compiler. But I still don't have a good test case to give to the Buble people to help them pinpoint the problem. Usually I just have to refactor/fiddle with code until it works right.

povrazor:
Actually if it's an ICE, then why Chrome is working?

Ah well. :smile:

StormBurpee:
Could it be differences between how Chromes Browser parses javascript, compared to Firefox?

povrazor:
Less talk, more sleep. :smile:

povrazor:
I had some other fixes to do to in the Calender, so I borrowed your 'objDate' line and that fixed it.

I made a tweak similar to what I usually do when fixing ICE's, but I didn't check before and after, so that might have done something too.

StormBurpee:
Rested and ready to see where I can help again!
Awesome, that's fine - so far it looks good, but I'll keep an eye out. 
Good Job with everything, PoV, you're doing an awesome job.

PR #342, opened 2016-12-02
Adds acceesskeys to slaughter page for easier theme slaughtering.

balohmatevz:
"a" for good
"s" for slaughter
"d" for bad

### DIFF
--- a/sandbox/public-theme/theme-slaughter.php
+++ b/sandbox/public-theme/theme-slaughter.php
@@ -92,13 +92,13 @@ function ShowSlaughter( $logged_in ) {
 //					"<div class='sg-item-x' onclick='kill_RemoveTheme("+Id+",\""+(ThemeAttr)+"\")'>❤</div>";
 				switch( value ) {
 				case 1:
-					node.innerHTML += "<div class='item-left item-good' title='Good'>✓</div>";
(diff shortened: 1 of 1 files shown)


povrazor:
I'm reluctant to merge this, as elsewhere the shortcut keys are the arrow keys (which are currently broken due to another bug).

balohmatevz:
While that's a better solution, this one can carry us through the period when the bug exists, as I'm sure there are other, much more pressing issues to fix than theme slaughter hotkeys. I added the accesskey to the code via the F12 console and it works great. 

So as a temporary fix, it works, and if it ends up being a permanent fix that's fine too, as having accesskeys is, while imperfect, still better than not having any hotkeys.

povrazor:
Thing is, I don't even use QWERTY. While WASD is good idea, they key choices are not international keyboard friendly.

balohmatevz:
Accept, slaughter, decline!

karai17:
What about multiple sets of keys? ASD + <v> + 123?

povrazor:
I'm not sure actionkeys (the feature proposed in the fix) supports more than 1 key.

balohmatevz:
I maintain the "It's better than nothing, even if imperfect" opinion.

povrazor:
Well come back to this later. #343

PR #358, opened 2016-12-03
Love

povrazor:


StormBurpee:
I'm unsure whether you've pushed this update to the live site, however when you're not logged in you can give love, refresh the page and give love.

povrazor:
Yes, it might sound weird but that's by design. That said, no matter how many times you do it, you can only give 1 love per IP, or per account.

PR #359, opened 2016-12-04
Temporary Countdown Clock in Sidebar

StormBurpee:
Just a temporary clock, that count downs to the date specified - in this case it's counting down to the Ludum Dare event. 
Just thought this might save you some time in the here and now, until you're ready to come back to this.


povrazor:
I'll take a look, but it might not work, as getElementById and similar functions don't work as expected with Preact's VDOM. Even still, I might be able to work with this to, like you said, get it going sooner.

StormBurpee:
Yes sorry - I've been meaning to do some reading up on Preact, and a few of the other frameworks. Just guess it gives a grounding!

povrazor:
No worries. You're efforts are much appreciated. I'll probably still find a use for this (to get me started if nothing else).

PR #440, opened 2016-12-08
Updated the setup script's database creation

gholdampf:
Pointed setup.sh to new database creation script, I don't think the old one was doing anything useful... now the database will be recreated whenever vagrant creates the VM. (I feel like the desired behaviour might be to keep the DB, and have the user manually update it if needed).


povrazor:
Cool cool. I'm not going to merge these in just (it's too close to LD). But when things cool down next week, I will.

Issue #459, opened 2016-12-08
Title on main page is ldjam.com


pta2002:
It should probably say 'Ludum Dare'

ComputerOverlord:
Sounds like an easy HTML fix..

zwrawr:
The Content (users, pages and post) create there title via their nodes name. ldjam.com is the root nod and so has no name, so it fall back to the web address.

This bit in content probably needs changing, [content.js#L25](https://github.com/ludumdare/ludumdare/blob/master/src/com/view-content/content.js#L25).

zwrawr:
Changing it to fallback to "Ludum Dare" seems reasonable as no other nodes should have empty names.

povrazor:
Ludum Dare's **true** root node is not the root node right now. Symlinks aren't working correctly yet, so I defaulted all sites to the common root (but both Jammer's are disabled so most people didn't likely notice).

PR #461, opened 2016-12-08
Changes FallBack Title To "Ludum Dare"

zwrawr:
Changes fallback title to "Ludum Dare" from the web address as suggested in #459 .


zwrawr:
When the ldjam.com's root node is set correctly to the ludumdare group(or event) then it wont have an empty node name and wont fallback anymore.

Issue #486, opened 2016-12-10
Embeds: Add a "open in new window" option


povrazor:
The TV widget is the reference for how embeds will soon work (and look).

![streamywidg](https://cloud.githubusercontent.com/assets/1570313/21074516/fc7ff6fe-bec8-11e6-8163-0e84a58442d6.png)

Instead of immediately embedding iFrame's of content, wrap a thumbnial with a play button.

What this issue is discussing is adding a button for opening the content directly in a new window. Similar to how Slack handles this, but I dislike how they made this a focal item right in the center (it should be an alt item). 99% of the time, people want to just play it, and when a browser is doing it's job, this is safe.

![slackers](https://cloud.githubusercontent.com/assets/1570313/21074511/d673290e-bec8-11e6-853b-73f922bbdfdf.png)

In the case of the TV widget though, I'm not sure where to put it, as the layout is populated with useful metrics (whose watching, etc). Maybe where the streamer name (and service logo) are?

References: #478

povrazor:
There is an example of this on the widget.

Issue #285, opened 2016-11-27
UX: No logout button


blinry:
A way to log the current user out should be exposed in the interface.

povrazor:
Yep! I haven't had time to finish the drop-down boxes that pop up with Notifications and when you click your Avatar. They go under the Avatar one.

tanrax:
thx

pfefferminzz:
First anniversary of this issue is tomorrow. If it's annoying to add a dropdown menu, the logout button could go here for now:

![ld-logout](https://user-images.githubusercontent.com/5459440/33243934-e47f6360-d2ee-11e7-8681-7a17ee12071e.png)

povrazor:
This got lost in a commit that was never pushed.

zwrawr:
Bump again

zwrawr:
It only took a year!

povrazor:
:fireworks:

pfefferminzz:
Thank you very much!

PR #537, opened 2017-04-02
Open Externally button - Issue #486

StormBurpee:
Adds a button next to the stream info in the bottom, for opening in a new tab
Proposes a fix for #486 

image

### DIFF
--- a/src/com/content-common/common-body-avatar.js
+++ b/src/com/content-common/common-body-avatar.js
@@ -1,0 +1,21 @@
+import { h, Component } 				from 'preact/preact';
+import { shallowDiff }	 				from 'shallow-compare/index';
+
+import NavLink							from 'com/nav-link/link';
+import IMG2								from 'com/img2/img2';
+
+export default class ContentCommonBodyAvatar extends Component {
+	constructor( props ) {
+		super(props);
+	}
+
+	shouldComponentUpdate( nextProps ) {
+		return shallowDiff(this.props.children, nextProps.children);
+	}
+
+	render( props ) {
+		var _class = "content-common-body -avatar" + (props.class ? " "+props.class : "");
+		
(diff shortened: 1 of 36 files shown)


StormBurpee:
Accidentally added a commit I meant for a separate pull request here... Forgot to create another branch sorry about that, but this was in terms of adding a temporary logout button, until you did the whole drop down thing for the user. Issue #285

povrazor:
I think the 'open externally' button needs to look distinct. The rounded look is the same as the metrics, which are just metrics.

I'm not sure what I was thinking when I made my notes in #486, but what I would probably do is make the button invisible, but when you're hovering over the TV widget, just as play button lights up, a secondary icon becomes visible in the lower right corner (`/www/icons/icomoon/enlarge.svg'). If you then choose to hover over that icon, that lights up instead of the play button.

With the words "Open Externally", it looks somewhat cluttered. Opening externally shouldn't be a common thing, but like #486 is something that's desirable design wise for embeds.

povrazor:
Also, while I appreciate the whitespace fixes, the shear amount of them makes it tricky to see which changes are whitespace and which are the proposed fixes.

povrazor:
As for the temporary logout button, the problem is the nav bar gets cluttered on mobile. So yes a temporary button can work, but once I let people sign up for the upcoming event, certain mobiles will show 4 icons wide and no longer fit.

Is it better than nothing?... Ha, I can't decide. :smile:

That's actually part of the reason put `<DropDownUser>` down for now, as though I could easily make a hover menu, I needed something that could work on Mobile (i.e. unrolled needs to be state, not parent hover trickery). 

I like that you changed `Fetch.Post` to a `Fetch.Get`, but I think `Fetch.Get` omits credentials by default ... or at least it should (if `{}` evaluates to true, then it's working by accident). Off the top of my head, the `{}` should be replaced with a `true`, but I would double check `src/shrub/internal/fetch` (or wherever it is... something like that).

PR #542, opened 2017-04-04
Fixed image overlap and sizing 

StormBurpee:
This is better now
image
image

### DIFF
--- a/src/com/sidebar-tv/tv.less
+++ b/src/com/sidebar-tv/tv.less
@@ -128,7 +128,6 @@
 	& > .-detail {
 		/*margin-bottom:0.5em;*/
 
-		margin-top: 1em;
 		line-height:1.25em;
 		height:2.5em;
 		overflow: hidden;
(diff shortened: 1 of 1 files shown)


povrazor:
Looks like the height is wrong. Notice how far away the Twitch Name+Icon is from the bottom in your examples above, compared to this:

fitted

StormBurpee:
image
image

StormBurpee:
actually

povrazor:
yeah

StormBurpee:
The detail is too far down

povrazor:
We're on the same wavelength now. :smile:

StormBurpee:
image
Does that look better?
I don't know, that right quote looks off to me, but I swear I haven't touched it

StormBurpee:
Oh, is the height not dynamic and you've just set it to two lines high?

StormBurpee:
Which, interestingly depending on how you want that part to be designed, could be brought up in an issue?

povrazor:
Yeah, it's max 2 lines of description/title, since one was never enough, and some people tend not to have restraint.

StormBurpee:
Well regardless, it should be fixed now!

povrazor:
Oh interesting, `calc`. I'm guessing that was the reason you were pleased about IE 11?

StormBurpee:
Exactly!

PR #554, opened 2017-04-09
fixed user page generation. emitting 401 errors when editing pages no…

povrazor:
…t yours


michcioperz:
I think it should not emit a 401 error, because 401 Unauthorized means that the user has not authenticated and the server has no way of knowing if the user should be able to do this, instead this should be a 403 Forbidden, because the user is authenticated, but not allowed to do this.
[Source](https://en.wikipedia.org/wiki/HTTP_403#Difference_from_status_.22401_Unauthorized.22)

PR #584, opened 2017-04-13
Added styling to markdown tables

zwrawr:
Currently markdown tables look like this.

This pr changes them to look more like the tables on github

### DIFF
--- a/src/com/content-common/common-body-markup.less
+++ b/src/com/content-common/common-body-markup.less
@@ -49,22 +49,24 @@
     /* Tables */
 
     table{
-        border: 1px solid #bbb;
+        border: 1px solid @COL_NLLL;
         margin: 2% 0;
         border-collapse: collapse;
     }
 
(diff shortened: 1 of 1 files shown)


povrazor:
I would suggest using the monochromatic colors from the palette (i.e. `@COL_N`, `@COL_NDDD`, `@COL_NLL`) instead of values like `#bbb`.

The color reference can be found here:

http://ludumdare.org/palette

The logic behind the names is that `COL_N` is the center gray, and the number of L's or D's says how many steps Lighter or Darker it is. 5 L's or D's are equivalent to the 'white' and 'black' (they're actually off white and off black, but those are the tones used).

If the color codes aren't working, then the file needs an `@import 'defs.less';` at the top.

zwrawr:
Now it's using the defines

povrazor:
Great. Merged. :smile:

PR #607, opened 2017-04-14
Dark code styling

### LINKED ISSUE #587: Syntax Highlighting: Make New (Dark) Theme
I'll be honest, I can't for the life of me remember why the source blocks use that gawd-awful yellow.

coolbug

Okay! So a lot of us are programmers here. While this isn't true for all programmers, a popular aesthetic of coders is the _dark theme_ in your source editor. i.e. something like this.

darktheme

**NOTE:** Do not copy this exactly. This is a random image I googled for illustrative purposes.

The current theme can be found here.

`/src/custom/prism/prism.css`

I would like someone to create a new dark theme using our palette.

http://ludumdare.org/palette

i.e. `COL_A`, `COL_B`, `COL_C` are the primary colors. `COL_AB` and its ilk are inbetweens. `COL_N` is the neutral monochromatic, and the variants like `COL_NDD` and `COL_NLLLL` are steps Darker and Lighter about the neutral monochromatic. 5 steps from `COL_N` is the "white" and "black" used, also known as `COL_L` and `COL_D`. True white and black are `COL_W` and `COL_K`, but shouldn't typically be used.

If you create a `.less` file (instead of `.css`), you can use the color codes listed in the palette.

```css
@import 'defs.less';

.stuff {
  color: @COL_A;
}
```

The code blocks should look decent when placed in both posts and comments.

monochrom

zwrawr:
PR to address #587 

Closes #587


povrazor:
Oh I like this. 

By the way, what happened to that C code shot?

povrazor:
Looks good. I like that you noticed you just had to _underscore a file to disable it. Merging it in.

zwrawr:
My local branch was just old. Didn't have the <tags> fix in it when i took the screen shot.

povrazor:
Okay. Just checking. :+1:

Jezzamonn:
This looks sooo good!! Nice one!

Probably not super necessary at this stage but still cool 😛

Issue #499, opened 2016-12-17
Toolchain: Generate a bunch of useful nodes


mikekasprzak:
To make it easier to get started, generate a bunch of useful nodes for people using the toolchain.

Examples:

* A user (or two)
* Some news posts
* an event (or two, in different states)

etc

povrazor:
This is sort of #491, but more explicit.

spirulence:
Combining a little research with a conversation with Mike on gitter:
> Sample posts, sample news posts, sample games, sample users, etc.

The extra script should create a few of each of these:

- [ ] posts
- [ ] news posts
- [ ] games
- [ ] users
- [ ] events
- [ ] tags

mikekasprzak:
Calling this done thanks to the simulator (though it needs to be better explained).

PR #611, opened 2017-04-15
Sort user's theme votes by time (#610)

### LINKED ISSUE #610: Theme voting - Sorting user's votes by time
The votes displayed on the theme slaughter page are indeed my votes, but they are not in the correct order (as added by JavaScript when voting).

I think the fix is just a `ORDER BY timestamp DESC` here:
https://github.com/ludumdare/ludumdare/blob/master/src/shrub/src/theme/theme_idea_vote.php#L11

(I set up the work environment but I'm not sure if there is a "create test data" command for the database, so I couldn't test this)

Aurel300:
Should fix #610 (although it is not tested, because #499)


povrazor:
Unfortunately timestamp isn't indexed, but `id` is, so I made a similar change using that. 

However, the way we're populating the array returned by the API actually implicitly sorts them by the key.

```
function _db_GetPair( &$st ) {
	$result = $st->get_result();
	$ret = [];
	while ($row = $result->fetch_array(MYSQLI_NUM)) {
		$ret[$row[0]] = $row[1];  // this here
	}
	return $ret;
}
```
So for either sorting to work, the above code needs to be tweaked to append rather than implicitly insert sort.

povrazor:
```
		$ret[] = $row[1];
```
That's not going to work. `$row[0]` is necessary, because what's returned are `key` = `value` pairs, not an ordinary array. The keys are important.

Aurel300:
Then it can only be `$ret[] = [$ret[0], $ret[1]];` which I didn't want to do because it changes the return type a lot, then needs to be mapped in the `themeIdeaVote_GetMy` …

povrazor:
I think what you might be looking for is `array_merge`, but be careful of this:

http://php.net/manual/en/function.array-merge.php

> Values in the input array with numeric keys will be renumbered with incrementing keys starting from zero in the result array.

That said, JSON always uses strings for keys, so they're being cast to strings before emitting the JSON anyway. So the potential fix for this might be to convert the key to a string while inserting.

That also brings up the question of whether the original syntax can be tweaked.

```
$ret[(string)$key] = $value;
```

I would recommend doing some tests with PHP, 'cause I'm not sure of the exact behavior (if insert always sorts, or that's just the case when keys are numbers).

Aurel300:
<s>I don't think you can have an array in PHP with numeric keys in an arbitrary order.</s>

This:

```php
    $arr = [];
    $arr[2] = "foo";
    $arr[3] = "bar";
    $arr[1] = "boo";
    foreach ($arr as $k => $v){
      echo $k.": ".$v."\n";
    }
```

Produces (running PHP 7.1.3):

```
    2: foo
    3: bar
    1: boo
```

Implying the array is not implicitly sorted by inserting numeric keys. So `db_QueryFetchIntPair` should technically work, but it doesn't?

PHP is weird.

povrazor:
Hmm. Is it being caused by the JSON encoder maybe?

```
json_encode($data, JSON_PRETTY_PRINT);
```

Aurel300:
Nope, produces:

```
{
    "2": "foo",
    "3": "bar",
    "1": "boo"
}
```

povrazor:
Ha, okay. Then I actually don't know what's going on. :smile:

povrazor:
Oh gosh! I'm fell for it! It's actually Chrome that's sorting, when you look at the preview view.

Aurel300:
Oh, it's sending every single theme to the user every time? That's not very nice.

povrazor:
Not every single time. Just on page load. It might not be cached at the moment, but eventually it could be.

povrazor:
Well that's embarrassing. I think your fix might have actually worked as is. :smile:

povrazor:
Okay I tested ASC vs DESC and saw no difference. It looks like something is happening on the JavaScript side (possibly another auto-sorting).

Aurel300:
This should finally do it?

povrazor:
I will get back to this, hopefully next week. Thanks for spearheading this though. :smile:

PR #654, opened 2017-04-18
Remove word-wrap: break-word; on body text

windsurfer:
word-wrap: break-word doesn't exist anymore, and should be changed to overflow-wrap: break-word (which was already in the CSS and was being overwritten by word-break: break-all). This fixes the odd wrapping behaviour:
screen shot 2017-04-18 at 13 47 32


povrazor:
Okay, we'll give this a try. We'll see if people can still break the layout after the change.

PR #663, opened 2017-04-18
See who your following

zwrawr:
`/users/<username>/following` not shows you who your following.

There's now a following button next to the feed button.
image

When you click it you get this, friends are sorted to the top of the list.
image
image

If you look at another person following page you get.
image

### TODO
- [x] Link usernames
- [x] Disable following button if you wont be able to see them
- [x] Fix user avatar styling 
- [x] Merge in upstream
- [x] Remove my avatar
- [x] rename ContentFollowing to ContentUserFollowing (as only users follow)
- [x] move `src/com/content-following/following.js` to `src/com/content-user/user-following.js`
- [x] don't use a loading or loaded state. Instead, initialize the state member holding your list to `null`. And when things are loaded, your call to `this.setState` will change it to something other than `null`, and trigger a re-render. The check becomes simple, an `if ( state.followingNodes ) {` which will be true if it contains your data.
- [x] modify the function definition to be `render( props, state )`, then use state instead of this.state.
- [x] move any functions that may trigger a `this.setState` out of the constructor and in to `componentDidMount() {}` (i.e. this.getFollowing)
- [x] move `src/com/content-userbar/userbar.js` to `src/com/content-user/user-bar.js` (and its .less file)
- [x] rename `.content-userbar` css style to `.content-user-bar`
- [x] rename the child style `.userbar` to `.-bar`, and tweak it to be a child of `.content-user-bar` by using `&`
- [x] Always make sure you have an `&` at the beginning of any child css style.
- [x] remove `console.log` from `src/com/view-content/content.js`
- [x] change `<div class="-bar">` to `<div class="content-user-bar">`
- [x] change #content .content-userbar{ } to #content .content-user-bar {}
- [x] remove `.-bar style` and `& .-bar {` line (and its `}`
- [x] change from `state.allowed` to a `hasPermission()` function


zwrawr:
Still need to fix a couple of links.

povrazor:
Very cool. I saw your tweet earlier. 

Also I :heart: _Mike page_. :wink:

povrazor:
I would remove `_uploads/zwrawr/profile.svg`, but not until you're done. :smile:

zwrawr:
Im done with this now, unless theres any feed back. 
Adding a followers tab will be easy now. Just need to grab nodes with ids from refs

povrazor:
Great work! 

Here's a whole bunch of feedback. This might take you a bit, but if you're still interested in doing more, I actually have a good task for you next. :smile:

* rename `ContentFollowing` to `ContentUserFollowing` (as only users follow)
* move `src/com/content-following/following.js` to `src/com/content-user/user-following.js` (I've been doing a cleanup, collapsing of folders together, but there's still a lot of unused things still around)
* don't use a `loading` or `loaded` state. Instead, initialize the state member holding your list to `null`. And when things are loaded, your call to `this.setState` will change it to something other than null, and trigger a re-render. The check becomes simple, an `if ( state.followingNodes ) {` which will be true if it contains your data.
* if you're using state in the a render function, modify the function definition to be `render( props, state ) {`. This only applies to render functions though (it's a feature of Preact/React). You can then use `state` instead of `this.state`.
* move any functions that may trigger a `this.setState` out of the constructor and in to `componentDidMount() {}` (i.e. this.getFollowing)
* move `src/com/content-userbar/userbar.js` to `src/com/content-user/user-bar.js` (and its `.less` file)
* rename `.content-userbar` css style to `.content-user-bar`
* rename the child style `.userbar` to `.-bar`, and tweak it to be a child of `.content-user-bar` by using `&` (see next section)
* Always make sure you have an `&` at the beginning of any child css style. Relationships work as follows:
  * `& > .-this-style-is-only-the-immediate-children-of-the-parent-style`
  * `& .-this-style-is-any-element-beneath-the-parent-style`
  * `&.-this-style-is-the-combination-of-the-parent-style-and-this-one`
* That first form (with the `>`) is the preferred form. Use this whenever possible. However if you're styling something and you don't know how far away it'll be from the parent, then without `>` is fine.
* remove `console.log` from `src/com/view-content/content.js`

I'm using the term `-bar` to represent a minimalistic interpretation of content that fill the full width of `#content`. To contrast, `-box` will be a box shaped version that can fit more than one per line (i.e. when viewing games, we compact them in to boxes, and fit as many as we can in depending on how wide `#content` is. Only a few content types will need a `-box` style, but many will need a `-bar` style. We don't have this yet, but eventually I would like have a way to view the raw feed of nodes. That's where the `-bar`'s come in. When new users create accounts, publish games, articles, etc. 

The added permutations of `-bar`'s can wait. Like I said before, if you're interested in doing more, I'm probably going to have you start on a pretty important `-box` type. :grin:

povrazor:
Since this is better than the interface I have, I think I'm going to piggyback on this for the team building feature (at least for now). Basically, I'll send the team-leader to their own `following` page, and add a button beside each of their friends to add them to their current games team.

zwrawr:
Think thats all the issues you brought up fixed.

povrazor:
Great. Let me give it a quick look.

povrazor:
Almost there! 

* in `user-bar.js` the line `<div class="-bar">` should be ` <div class="content-user-bar">`. The class `ContentUserBar` should be emitting something with the class `content-user-bar`.
* in `user-bar.less` the line `#content .content-userbar{` should probably be `#content .content-user-bar {`
* it looks like the `.-bar` style is unnecessary, so I'd suggest deleting the `& .-bar {` line (and its `}`).

povrazor:
I would suggest getting rid of the `allowed` state too, and replacing it with a function. Something like `hasPermission()` or something like that (I'm not entirely sold on the name). Have it return the result of that comparison you do in the constructor.

povrazor:
Okay!

There a minor thing (the function declaration `hasPermission(props){` doesn't need props there, since you're pulling the props from `this.props`).

I think that's it. Have you tested it, and is it working as expected in your local build?

If so, I'll take this and push it to the live site.

zwrawr:
Fixed the props thing. It works on my local build.

povrazor:
Okay then. Lets push big green button...

povrazor:
Awesome awesome. Just tested it in my build.

hams

This is a great addition. Thanks for doing it, and putting up with all my requests. :smile:

povrazor:
And it's now live.

Thanks again for your hard work. :smile:

Like I said earlier, if have the time and you're still interested in doing more, I have a task for you. If you come to the Gitter chat I can brief you.

PR #668, opened 2017-04-19
Sort user's theme votes by time (#610)

Aurel300:
Sorry for closing and reopening, I forgot to put this in a branch.


povrazor:
Ah okay, I wasn't sure what had happened.

sgstair:
I can't tell what the status of this is from the related comment threads - But good news! We do now have a simulator to generate test data, and it's now easy to test this stuff manually in the dev environment. If this is still outstanding / waiting, take a look at [this page](https://github.com/ludumdare/ludumdare/wiki/Adding-a-new-Event) for how to test things out.

povrazor:
Yeah this is a long time ago commit. If I remember correctly, I think I accidentally gave the wrong advice for fixing this issue, and was swamped at the time so it went untouched. I'm not sure what effect this will have, but it would be nice to test and see.

sgstair:
I looked closer at this; The fix still adds redundant DB functions, and doesn't actually guarantee the order it wants (sql querying doesn't guarantee things come back in ID order, so should order by to enforce that). I get the idea though, if you don't want to update the PR I may rework this change for the current codebase.

PR #677, opened 2017-04-19
Add guide and info subtype headers

Noojuno:
This

vs this


zwrawr:
That seems like a good idea. :+1:

Issue #580, opened 2017-04-12
Markup editor, Issues with protocol-less links.


zwrawr:
### Links with the protocol part work, such as the one below.
![](http://image.prntscr.com/image/a319d7c42a42405a80a23db28ce28934.png)
![](http://image.prntscr.com/image/810c4e78e7d64ac6995f576c86963b2e.png)

### But the preview fails if the protocol is omitted.
![](http://image.prntscr.com/image/8994740398004a299b739de2c0122c48.png)
![](http://image.prntscr.com/image/789d340c78ea4813a1e0a58e32adf01d.png)

## And generates the following error messages
```
all.min.js?v=2147-f2242a1:8 [constructor]
all.min.js?v=2147-f2242a1:8 [fetchData]
all.min.js?v=2147-f2242a1:8 [fetchUser]
all.min.js?v=2147-f2242a1:8 [fetchRoot]
all.min.js?v=2147-f2242a1:8 [fetchRoot] Done: 1
all.min.js?v=2147-f2242a1:8 [fetchUser] caller_id: 29
all.min.js?v=2147-f2242a1:8 [fetchUser] Done: 29
all.min.js?v=2147-f2242a1:8 [fetchNode]
all.min.js?v=2147-f2242a1:8 [fetchNode] Done: 29
all.min.js?v=2147-f2242a1:4 Potentially unhandled rejection [1] TypeError: Cannot read property 'split' of undefined
Please report this to https://github.com/chjj/marked.
    at e.extractFromURL (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:28242)
    at e.hasEmbed (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:30071)
    at n.link (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:5367)
    at t.outputLink (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:3049)
    at t.output (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:1610)
    at o.tok (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:7815)
    at o.parse (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:5991)
    at Function.o.parse (https://ldjam.com/-/all.min.js?v=2147-f2242a1:3:5839)
    at Function.f [as parse] (https://ldjam.com/-/all.min.js?v=2147-f2242a1:2:25951)
[...]

PR #693, opened 2017-04-20
Fix [](www.example.com) links

zwrawr:
## Markdown
```md

##### zakwest
https://www.zakwest.tech  
[Markdown Test Link](https://www.zakwest.tech)  
[](https://www.zakwest.tech)

```

## Old
image

## New
image


zwrawr:
There a second problem here. Its to do with protocol less links #580

PR #717, opened 2017-04-21
Wrapper for youtube embeds

zwrawr:
## Not hovering
image

## Hovering over thumbnail
image

## Hovering over external
image

## After clicking play
image


zwrawr:
Addresses #486 only for youtube, but then again that's the only thing were embedding

PR #718, opened 2017-04-21
Love system for notes

Noojuno:


povrazor:
I've merged this already (since it does work), but there are some bugs.

* This might sound strange, but you actually need to be allowed to give 's while not logged in. I.e. user.id is zero. In that case it's associated with the IP address, rather than the account.
* Also it looks like if I have 2 tabs open, I remove it in one tab, try to remove it again in the other tab it's stuck on.

Not a showstopper, but they should be addressed.

Issue #748, opened 2017-04-23
Spinner is not removed in the user page


JortHub:
When my user page is loading the spinner doesn't stop. In the Network tools Chrome says the webpage isn't loading anything and the page looks like it's fully loaded. I have reloaded the site a few times, but the spinner still stays on the screen. The spinner also remains when I emulate other browsers.

A printscreen of the site:
![ld](https://cloud.githubusercontent.com/assets/10116429/25312754/3d204832-2821-11e7-95e5-7165663ebc7f.png)

This might also be useful:
![ld2](https://cloud.githubusercontent.com/assets/10116429/25312793/e85685fe-2821-11e7-9a32-318c822f2cbb.png)

And a profile when loading the site:
[Profile.zip](https://github.com/ludumdare/ludumdare/files/949619/Profile.zip)

**Browser:** Chrome (57.0.2987.133)
**OS:** Windows 10 1(0.0.14393)

zwrawr:
Okay i think i know what the problem is. What happens if you click the "join event" button and make a draft of your game submission.

JortHub:
When I create a game it is removed

zwrawr:
Can you click this link and paste the results here https://api.ldjam.com/vx/node/get/9420?pretty

JortHub:
Sure:
`{
    "status": 200,
    "caller_id": 9420,
    "node": [
        {
            "id": 9420,
            "parent": 2,
            "superparent": 0,
            "author": 9420,
            "type": "user",
            "subtype": "",
            "subsubtype": "",
            "published": "2017-03-27T18:13:49Z",
            "created": "2017-03-27T18:13:49Z",
            "modified": "2017-03-27T18:13:49Z",
            "version": 9396,
            "slug": "jortvd",
            "name": "JortvD",
            "body": "",
            "path": "/users/jortvd",
            "parents": [
                1,
                2
            ],
            "games": 1,
            "articles": 0,
            "posts": 0,
            "meta": [],
            "link": [],
            "love": 0
        }
    ]
}`

zwrawr:
Looks to be a similar issue, we need to add a check to make sure you have posts before we try to show them. In the mean time if you make a post then your feed should work.

zwrawr:
Okay, i can reproduce this with a new account, http://prntscr.com/eztex8 Will  get to work on a fix.

zwrawr:
@JortHub is this still an issue for you? I think this is fixed

[...]

Issue #747, opened 2017-04-23
Cannot see users I am following


gonutz:
I do the following:
- click on the user icon in the top-right
- click on the 'Following' button
Below the user description the spinning circle appears, indicating that the list is loading, but it keeps spinning for ever.

zwrawr:
Can you tell us what browser and os your using. and can you see if there are any errors in the javascript console. (In chrome press `Fn12` (or right click and choose `inspect`) and click on the `console` tab.

dickclaus:
Hey, I'm using latest Firefox on MacOS. Have the same problem. Here is log from console.

```
GET 
XHR 
https://api.ldjam.com/vx/node/walk/1/users/dick-claus/following [HTTP/2.0 200 OK 151ms]
GET 
XHR 
https://api.ldjam.com/vx/node/get/12583 [HTTP/2.0 200 OK 142ms]
[fetchNode] Done: 12583  all.min.js:8:28064
GET 
XHR 
https://api.ldjam.com/vx/node/get/11+10536+4287+5667+2584 [HTTP/2.0 200 OK 214ms]
 Potentially unhandled rejection [1] ContentUserFollowing</t.prototype.render/n<@https://ldjam.com/-/all.min.js?v=2470-668569c:7:25936
ContentUserFollowing</t.prototype.render@https://ldjam.com/-/all.min.js?v=2470-668569c:7:25896
renderComponent@https://ldjam.com/-/all.min.js?v=2470-668569c:1:7713
rerender@https://ldjam.com/-/all.min.js?v=2470-668569c:1:1392
F@https://ldjam.com/-/all.min.js?v=2470-668569c:4:24558
L@https://ldjam.com/-/all.min.js?v=2470-668569c:4:24192
[7]</</</</x.prototype.when@https://ldjam.com/-/all.min.js?v=2470-668569c:4:27863
[7]</</</</I.prototype.run@https://ldjam.com/-/all.min.js?v=2470-668569c:4:28952
[3]</</</e.prototype._drain@https://ldjam.com/-/all.min.js?v=2470-668569c:4:18805
e/this.drain@https://ldjam.com/-/all.min.js?v=2470-668569c:4:18472
t@https://ldjam.com/-/all.min.js?v=2470-668569c:4:20395
```

gonutz:
I am using Firefox 53 on Windows 8.1.
Can you tell me how to get that log in Firefox?

zwrawr:
[...]

PR #759, opened 2017-04-23
More No Content Fixes

zwrawr:
Stops "No Games" (and similar) messages flashing up whilst the content is being loaded. Now they only come up if we have tried to load content but got none


zwrawr:
More stuff for #747 and #748

Issue #809, opened 2017-04-28
Markdown: Add "spoilers" support


mikekasprzak:
> Local Minimum‏ @Local_Minimum
>
> what markdown engine does ldjam use/can it be made to support spoilers? hidden text that show with user interaction.
> Would be awesome on game submit & comment to give extra instructions that may spoil content or discuss content more freely

povrazor:
I don't know what other markdown engines use for spoilers, but I quite like this idea.

zwrawr:
Totally possible.
Its made trickier by the fact that were not using es6 for the markdown engine. But the youtube embed can be used as an example on how to switch out content based on user interaction.

povrazor:
I found one mention of Spoiler syntax. It works similar to blockquotes.

```
!> This text is a spoiler
!> Don't spoil me bro
```
http://stackoverflow.com/a/32183830/5678759

povrazor:
Another spoiler syntax looks like this.

```
>! this is a spoiler
> ! this is also a spoiler
```
https://meta.stackexchange.com/a/72878

The link above demonstrates it. I don't think that I like that hovering reveals it (it's mobile friendly).

I'm not entirely sure what I think of this alternative (i.e. `!` after the blockquote). I think I prefer the syntax with it before the blockquote, so everything on the block quoted line can just be assumed to be in the block.

local-minimum:
I saw that linestart `!>` right before falling asleep yesterday and it seems reasonable and doable. That you can do other formatting is important, because you may want to say something like

```
!> Pressing `X` on the tree trunk will make the squirrel throw a nut on you
```

Checked the marked.js code and I almost partially think I understand how it would be done on the js-side. Then it's just css for the effect, I guess?

Inline spoiling isn't really as needed, and could be added at a later date if people request it by some similar styled `!!>spoiler!!>` thingy.

local-minimum:
I made what I think are the needed changes here:

https://github.com/ludumdare/ludumdare/compare/master...local-minimum:patch-1

But I don't know how to test them really (and since there's no css to go with it, they won't really have an effect..

Should I pull request them?

zwrawr:
Definitely PR them! If you want to style them then [here](/ludumdare/ludumdare/blob/master/src/com/content-common/common-body-markup.less) would be a good place to do it.

local-minimum:
That CSS was above my skills but I imagine something like start with black text och black background and then default to common style when clicked on or something.

[...]

Issue #801, opened 2017-04-26
@gmail.com email address links to non-existent user page for "/users/gmail"


ReactorScram:
https://i.imgur.com/sPhhBmD.png

This is in the Markdown description for my game. 

The site sees "username@gmail.com" and turns "@gmail" into a link to an imaginary ldjam user.

Hope this is the right place to report this.

zwrawr:
i swear we already fixed that

povrazor:
I have pending changes with the markdown parser. Let me get those in first.

povrazor:
Okay. This one is fair game.

Aurel300:
I mean currently in `custom/marked/marked.js` lines 465 and 890 are commented out, is there any reason an e-mail would ever render?

povrazor:
I disabled e-mails a long time when it was causing problems with `@names` (more important IMO). There's has been a bunch of work in those files since, and fixing it was definitely talked about, but either it got shelved, or any other changes it was associated with got thrown away.

So yeah, at least as far as I now automatic e-mails are unsupported. They may work if you make a link by hand (i.e. `[e-mail me](mark@bark.fark)` or somesuch). I haven't checked though.

There's also the issue of e-mail obfuscation.

I've recently discovered that leaving an e-mail in the open is a magnet for garbage. Over on `ludumdare.com`, I also have my e-mail in the open, but it's wrapped by some clever WordPress plugin that obfuscates it somehow. My best guess is it doesn't show up as an e-mail when you crawl the page, but gets converted to one by javascript (a quick decrypt or something).

Then again, our entire site is now generated by JavaScript now. So there's no such thing as e-mails in the open. It's not encrypted or anything, but to farm an address requires an advanced bot with a working JavaScript parser.

zwrawr:
Links in the form
`[my email](noreply@ldjam.com)` create links to `ldjam.com`

local-minimum:
From StackOverflow http://stackoverflow.com/a/46181/1099682 

The automagic inline mail regex should probably be:

```
 email:  /^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i,
```

but it requires special version for where it is executed, as captures partials too after the full one.

povrazor:
#233

local-minimum:
[...]

PR #815, opened 2017-04-29
Added spoiler markdown

local-minimum:
This should hopefully convert block

```
!>Some important spoiler
```
into
```
<span class="spoiler">Some important spoiler</span>
```

It is fully untested as I don't know how that works.

### DIFF
--- a/src/custom/marked/marked.js
+++ b/src/custom/marked/marked.js
@@ -84,7 +84,7 @@ block.normal = merge({}, block);
 block.gfm = merge({}, block.normal, {
   fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\s*\1 *(?:\n+|$)/,
   paragraph: /^/,
-  heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/,
+  heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
 });
 
 block.gfm.paragraph = replace(block.paragraph)
--- a/src/custom/marked/marked.js
+++ b/src/custom/marked/marked.js
@@ -16,7 +16,7 @@ var block = {
   fences: noop,
   hr: /^( *[-*_]){3,} *(?:\n+|$)/,
   heading: /^ *(#{1,6}) *([^\n]+?) *#* *(?:\n+|$)/,
-  spoiler: /^(\!>)([^\n]+?) *\n*/,
+  spoiler: /^( *\!>[^\n]+(\n(?!def)[^\n]+)*\n*)+/,
   nptable: noop,
   lheading: /^([^\n]+)\n *(=|-){2,} *(?:\n+|$)/,
   blockquote: /^( *>[^\n]+(\n(?!def)[^\n]+)*\n*)+/,
@@ -43,6 +43,11 @@ block.list = replace(block.list)
 block.blockquote = replace(block.blockquote)
   ('def', block.def)
   ();
+
+block.spoiler = replace(block.spoiler)
+  ('def', block.def)
+  ();
+  
 
 block._tag = '(?!(?:'
   + 'a|em|strong|small|s|cite|q|dfn|abbr|data|time|code'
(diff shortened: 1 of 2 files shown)


local-minimum:
It should be noted that this will treat each line starting with `!>` as a separate spoiler (if I understand things correctly). But I think that should be a good enough first iteration

povrazor:
I'm reluctant to add this untested, because the markdown parser is perhaps the most finicky piece of code in the entire codebase.

Also you should be basing this off the `blockquote`, as the feature is effectively identical (just you'll want special `<div>` to wrap it instead of `<blockquote>`).

local-minimum:
It sounds reasonable to be scared of adding it untested.

I'll get working on it, but it will probably take a while with playing ludum games and making a game for an other jam in parallel.

This is work on (put here to get link) #809

zwrawr:
It also worth noting that the part of the markdown parser that is a pain is the linking, i don't think there been any issues with the rest of it. 

So don't be get scared of the evil markdown parser, the stuff your messing with shouldn't break it.

povrazor:
Well there's a reason e-mails don't work anymore. ;)

local-minimum:
I agree with Mike, it's better to test.
With regards to email, the regex seems wrong (doesn't allow for most mails). Should I fix that here too? And do you have an issue where this is discussed?

zwrawr:
A separate PR would be better.
#801 is about messed up email hadling

local-minimum:
spoilersimple

Simple non-fancy support tested and works. Will not spoil formatted stuff like links and doesn't have any nice effects on hover or click because I don't really understand `less` nor found any reasonable introduction.

zwrawr:
Less is just some syntactical sugar on top of css. If you just write css in less it'll work.

The main thing is that less like this
```
@blue: #00f;

#content{
     div{
        color: @blue;
     }
     img{
        width: 100px;
     }
}
```

is turned into css like this
```
 #content div{
    color: #00f;
 }
 #content img{
    width: 100px;
 }
```

local-minimum:
spoilerv2

This is how far I think I can take it. I don't understand why some urls won't get spoiled while others do.
Tested on FireFox

src/custom/marked/marked.js
```diff
@@ -77,7 +84,7 @@ block.normal = merge({}, block);
 block.gfm = merge({}, block.normal, {
   fences: /^ *(`{3,}|~{3,})[ \.]*(\S+)? *\n([\s\S]*?)\s*\1 *(?:\n+|$)/,
   paragraph: /^/,
-  heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/
+  heading: /^ *(#{1,6}) +([^\n]+?) *#* *(?:\n+|$)/,
```
zwrawr: Is that trailing comma meant to be there?

local-minimum: I'll fix, that was where the previous spoiler implementations went.

local-minimum:
Just so that I know, is there anything more I should do with this or just wait for LD to be over and things settle down?

zwrawr:
Basically wait for pov,
He's probably burnt out after LD and taking a break.
You should take a look at @Noojuno Pull Request #846 . It is a massive rework of the whole markdown and embeding code. You too a definitely going to conflict, so one of you will have to integrate your changes with theirs.

local-minimum:
While a bit frustrating, the code in #846 is much nicer organized than what is in master and would have been easier to develop and debug in. We'll see what pov and @Noojuno says and wait.

povrazor:
Sorry, I'm caught up with some house cleaning today. I'll hopefully get some time tonight to tackle some of the outstanding PR's.

zwrawr:
Yea, it's bad timing that the @Noojuno finished the markdown re work soo fast. I expected it to take way longer to re write marked to be es6 complaint.

local-minimum:
Really you should take a break pov and not stress this at all, just wanted to be sure it wasn't expected to deal with the strange links that appear in the spoiler at times.

zwrawr:
This merge conflict is going to be fun to fix. If you get stuck jump in the chat https://gitter.im/ludumdare/ludumdare.

PR #821, opened 2017-04-30
Prevent pictures from overflowing the comment box

mkalam-alami:
An issue regarding big pictures in comments was noticed [by SirRucTheRandom](https://ldjam.com/events/ludum-dare/38/tub-of-war) (see one of the last comments).

A little CSS tweak fixed it:

fixed


povrazor:
Thanks for the fix. That did the trick.

PR #851, opened 2017-05-04
Brand Colors, 

zwrawr:
Addresses #144 

Since @Noojuno has made it way easier to mess with the auto-embed code, were probably going to start adding more embeds. Were going to want to add logo's of these sites and style them with the appropriate branding colours. As mentioned in #144 we really don't want to have to go out and find all brand colours ourselves. This is where https://brandcolors.net/ comes in. Unfortunately were dealing with a lot of gaming and development specific sites which aren't included. In that case we have to add the colours our-self.

This commit adds.
* https://brandcolors.net/ style sheet `src/external/BrandColors/brandcolors.less`
* A readme with details with brandcolor site and issues page. *(note: there is no licence on the .less file, probably because all the colors are 'owned' by the brands them selves)*
* Internal extra site colors (itch. gamejolt, 1h game jam) `src/internal/BrandColors/brandcolors.less`
* Makes Youtube logo use the youtube color image


zwrawr:
Fixed merge conflicts and slightly changed styles.

Here it is when hovering over the open externally.
image

sgstair:
These old PRs have been stalled for a while, is there still interest in getting them merged?

povrazor:
This is fine, but instead of splitting it between `/internal` and `/external`, both should be in `/custom`.

Also the custom names should match the coding style of the original API (i.e. `@bc-itch-io`).

Finally there should again be a singe include. I like that our additions are in a separate file (something like `customcolor.less` instead of `brandcolors.less`). Either add an include at the top of `brandcolors.less` for our customs, or something better (if you think of something better).

povrazor:
Also I don't like that it's `.youtube`. Please prefix it with a dash (`.-youtube`) to mean a child style, as per RSCSS: https://github.com/ludumdare/ludumdare/wiki/Style-Guide#css-coding-style

zwrawr:
Killing this in favor of #1260

PR #856, opened 2017-05-07
[WIP] LD Local Links

zwrawr:
# LD local links
Make local links look good and not cause reloads
image

Somewhere in the gitter chat mentions how `//` and `///` Links are meant to work but i'll have to look back to find out. Leaving this for a while till I get more time.

## TODO 
* Get `//` links working. (I think these should work the same as `/` links)
* Get `///` links working. Linking to images, but not as an image.
* Get relative links working
* Remove other logging


local-minimum:
This is probably just reflecting my understanding but why is there need for the extra uri-logic in the autoembed.js? I kind of though that what cont'd out of the marked render should be ready to use?

zwrawr:
Autoembed.js handles site specific embeds. E.g. youtube players or twitter embeds.

In this case its being used for ld specific link styling. The way were detecting internal links is kinda broken so i still need to clean it up.

local-minimum:
Yeah I think that the way there was stuff about links being internal there or not was confusing me. 

Anyhow , this relates to https://github.com/ludumdare/ludumdare/issues/879 I think.

povrazor:
Basic fix to solve merge conflicts. May trigger the linter since this code was previously repaired.

src/com/autoembed/autoembed.js
```diff
@@ -47,12 +53,20 @@ export default class AutoEmbed extends Component {
     }
 
     if (url.domain) {
+
+      /*
+      * Embeded content
+      */
       if (url.domain.indexOf('youtube.com') !== -1) {
         if ((url.paths && url.paths[0] == 'watch') && url.args.v) {
           return <YoutubeEmbed href={props.href} />;
```
povrazor: I realize this isn't from this PR (again, I'm taking notes for linter/style guide improvements), but prefer this:
```
if ( url.domain.indexOf('youtube.com') !== -1 ) {
          if ( url.paths && (url.paths[0] == 'watch') && url.args.v ) {
```
I mention spacing of special keywords elsewhere.

Of course use added brackets as needed, but a rule of thumb is to look at what's inside the conditional. 

* If it's a single conditional/implicit statement (i.e. `if ( url.paths && url.args.v ) {`) it's fine without added brackets
* If it's only a single `LHS == RHS` statement, it's fine without added brackets
* If multiple statements are used, separated by `&&` or `||`, and one is a `LHS == RHS` statement, use added brackets just to visually isolate it. i.e. `if ( other && (LHS == RHS) ) {`
* If the LHS or RHS side is more than a single item, use brackets to visually isolate the complexity. i.e. `if ( (a+b) == 12 ) {`

`&&` operations in series will always require the previous one to succeed, so no extra brackets are needed. When you mix and match `&&` with `||`, extra brackets will be necessary to guarantee order.

`&&`, `||`, assignment (`=`), and equality tests (`==`, `!==`, `>=`, etc) are special, and should be given extra spaces around them to make them easier to notice. 

Whether to space arithmetic operators (`+`, `-`, etc) is up to the person writing the math. If it's easier to follow with spaces, then use spaces.

src/com/autoembed/autoembed.js
```diff
@@ -77,9 +91,22 @@ export default class AutoEmbed extends Component {
         return (
           <SmartLink icon_name='soundcloud' full_url={url.href} domain={lit} part_url={unlit}></SmartLink>
         );
-      } else if ( url.domain.indexOf('//'+window.location.hostname) !== -1 ) {
-        return (<span class="smart-link"><a href={props.href}><strong class="-the-rest">{props.href}</strong></a></span>);
-			}
+      } else if (url.domain.indexOf(window.location.hostname) == 0 ) {
```
povrazor: Prefer
```
}
else if ( ... ) {
```

src/com/autoembed/autoembed.js
```diff
@@ -77,9 +91,22 @@ export default class AutoEmbed extends Component {
...
+        // local links starting pointing to the same domain name e.g. http://ldjam.com/users/mike
+        console.log(lit);
+        console.log(unlit);
+        console.log(url);
+        console.log(props);
```
povrazor: Try to remember to comment out or remove logging that is no longer necessary.

src/com/autoembed/autoembed.js
```diff
@@ -77,9 +91,22 @@ export default class AutoEmbed extends Component {
...
+
+        return (
+          <LocalLink href={url.href} text={(unlit)?url.path:props.text} title={''} target={"_blank"}/>
+        );
+      } else {
```
povrazor: Prefer
```
}
else {
```

src/com/autoembed/locallink.js
```diff
@@ -0,0 +1,27 @@
...
+
+import NavLink        from 'com/nav-link/link';
+import SVGIcon        from 'com/svg-icon/icon';
+
+export default class LocalLink extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render(props, {}) {
```
povrazor: Prefer
```
render( props, {} ) {
```

src/com/autoembed/smartlink.less
```diff
@@ -13,3 +15,31 @@
...
+                color: @COL_A;
+            }
+            &.icon-d-are{
+                color: @COL_AB;
+            }
+        }
+    }
+
+    &:hover,
+    &:active{
```
povrazor: Prefer
```
&:active {
```
(notice the space between active and `{`)

src/internal/autoembed/autoembed.js
```diff
@@ -40,20 +40,19 @@
...
 
-		// if its not already an external link with a protocol and it has a dot in it
-		// then make it an external link becuase not internal links have dots
-		if ( (ret.url.indexOf('//') == -1) && (ret.url.indexOf('.') != -1) && (ret.url.indexOf('/') == -1 || ret.url.indexOf('.') < ret.url.indexOf('/')) ) {
-			// Assume HTTP. Most HTTPS sites will autoredirect to HTTPS
-			ret.url = 'http://' + ret.url;
-		}
-		else if ( ret.url.indexOf('//') == -1 ) {
-			ret.internal = true;
+		if (ret.origin.indexOf('/') == 0 || ret.origin.indexOf('.') == -1) {
```
povrazor: Prefer 
```
if ( (ret.origin.indexOf('/') == 0) || (ret.origin.indexOf('.') == -1) ) {
```

src/internal/marked/Renderer.js
```diff
@@ -4,14 +4,193 @@ import Util 				from './Util';
...
 export default class Renderer {
-	constructor(options) {
-		this.options = options || {};
-	}
+  constructor(options) {
+    this.options = options || {};
+  }
+
+  code(code, lang, escaped) {
+    if (this.options.highlight) {
```
povrazor: Prefer `if ( ... ) {` here and below

src/internal/marked/Renderer.js
```diff
@@ -4,14 +4,193 @@ import Util 				from './Util';
...
+    return (
+      <blockquote>{quote}</blockquote>
+    );
+  };
+
+  html(html) {
+    return {html};
+  };
+
+  heading(text, level, raw) {
```
povrazor: prefer
```
heading( text, level, raw ) {
```
'cause it's a function definition

src/internal/marked/Renderer.js
```diff
@@ -4,14 +4,193 @@ import Util 				from './Util';
...
+      <BlockSpoiler>{secret}</BlockSpoiler>
+    );
+  };
+  blockquote(quote) {
+    return (
+      <blockquote>{quote}</blockquote>
+    );
+  };
+
+  html(html) {
```
povrazor: prefer
```
html( html ) {
```
because it's a function definition

zwrawr:
Closed in favour of #1293

PR #923, opened 2017-05-15
[Untested] We don't need that many decimal points

zwrawr:
image

This should dop it to 2 d.p.


zwrawr:
Wtf, So the online github editor has bugs.

povrazor:
Make it 3 decimal places. 2 isn't enough.

povrazor:
Acutally nevermind. I need to fix it anyway....

Issue #380, opened 2016-12-04
Resource Pack


karai17:
Would be nice if there was a resource pack for people who want to use Ludum Dare branding for streams or splash screens, etc. Most notably, an SVG/high res logo in various shapes and form factors.

povrazor:
Agreed. #194

povrazor:
For the time being, you can grab the unprocessed SVG files of the `Ludum` and `Dare` logo words here.

https://github.com/ludumdare/ludumdare/tree/master/src/icons/ludumdare

karai17:
I opened up those SVGs in Blender to make my own processed logo and I noticed that the glyphs weren't all at the same X base level. When normalizing all the bases to 0, the glyphs now look a bit off height-wise.

![image](https://cloud.githubusercontent.com/assets/590332/20894332/5590fe46-baeb-11e6-968a-3fe297c6f9a3.png)

povrazor:
Yes, the glyphs work best if you leave them. Half-circles do tend to droop more than flats-sides.

zwrawr:
Colored svg with a black background.

PR #958, opened 2017-05-19
Open Graph meta initial test

Noojuno:
In theory should make links from the site look like this on twitter and facebook, but due to the way the site is set up and it's async nature I'm not entirely sure. At the very least it won't break anything, but it needs to be live to really be tested.

### DIFF
--- a/src/com/view-content/content.js
+++ b/src/com/view-content/content.js
@@ -373,7 +373,7 @@ export default class ViewContent extends Component {
 			else if ( extra && extra.length && extra[0] == 'games' ) {
 				let DefaultSubFilter = 'all';
 				let DefaultFilter = 'grade';//'danger';//'smart';
-
+				
 				function EvalFilter(str) {
 					let MappingTable = {
 						'all': 'compo+jam+craft+release',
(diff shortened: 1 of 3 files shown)


local-minimum:
Just for me to understand the code better, where is `node.meta` set or is that for the future when this proof of concept is working to add that into suitable a place? The render-function?

zwrawr:
node.meta is from the node meta DB table,  In this case (for games) it contains cover images ( and opt-outs?)

povrazor:
Yeah, when you request a node from the server, I'm including a bunch of data, not just what's in the `sh_node` table.

local-minimum:
Alright so it's part of the json from the server, but as far as I can see there's only `meta.cover` on game-nodes and profile-nodes.

zwrawr:
Maybe make the default image something like the one at the bottom of this issue https://github.com/ludumdare/ludumdare/issues/380

povrazor:
Okay, lets try it. I'm not entirely sure it's going to work, but if it does hey.

povrazor:
Yeah, no effect. Metadata is set, but it's not tracked by Twitter. 

failes

PR #974, opened 2017-05-20
Nav root fix

local-minimum:
Currently if you are not logged in you don't get the `Feed` option in the nav-root.
The feed is available while not logged in.

This adds feed to nav-root independent on being logged in but keeps the previous defaulting behaviour.


povrazor:
I'm somewhat confused about what this is for.

local-minimum:
If you are not logged in you only see games abd news on the main site not the feed of what people have posted. This seems wrong to me .

povrazor:
There seems to be several other changes mixed in, and this breaks other behaviours.

Anyway, what you're asking for is easy. I've made the change.

PR #975, opened 2017-05-20
better games counting

local-minimum:
I think this will include games like they are on `/user/games` into the
count given as response to the `/user`, that used to decide if `Games` button
appeared.

__NOTE__: It builds and it can return 0 without an issue in my
test environment, how it works when there are games, I don't
know / need help setting up a fuller test environment some day
after the LD is over.

### DIFF
--- a/src/shrub/src/node/node_core.php
+++ b/src/shrub/src/node/node_core.php
@@ -336,7 +336,30 @@ function node_CountByAuthorType( $ids, $authors, $types = null, $subtypes = null
 //				GROUP BY id".($multi?';':' LIMIT 1;'),
 //				...$ARGS
 //			);
-	}
+			$QUERY[] = '`key`=?';
+			$ARGS[] = 'author';
+
+			$pre_query[] = 'b'.$node_query;
+			if ( isset($node_args) ) $ARGS[] = $node_args;
+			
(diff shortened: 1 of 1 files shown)


ajayyy:
Do we need to hide the games button?

povrazor:
I'm still waiting to hear back, but this is sounding awfully familiar.

https://twitter.com/DaFluffyPotato/status/893337787394662400

local-minimum:
i haven't managed to setup events and so i haven't managed to create games

sgstair:
It seems like this isn't necessary, because /node/getmy will already return you a list of games you're an author on. (refs.author). And it's not related to the tweet pov linked (which is a bug we already fixed at the very tail end of LD39)

povrazor:
Yeah my bad for not being more on the ball here.

povrazor:
Yeah I think this might have been something I solved with a 1-line fix. I seem to remember a bug that made games not show up on the game page if you didn't have any games, and doing a quick-fix change it to return some better total than what I was returning before. 

Sorry again. I think this might have been the result of an early communication problem, something I should have made more clear I was dealing with.

Or if it's not, haha, we can just keep talking about this issue until we get bored. :laughing:

local-minimum:
I was thinking about closing it, but in general to have defaults that would stress ppl of they ever show all seems like a bad idea to me

PR #979, opened 2017-05-20
Handling not geting votes data in more graceful way

local-minimum:
Can build, but I can't test given current state and tricky because server also needs to not respond with data.

### DIFF
--- a/src/com/content-item/item.js
+++ b/src/com/content-item/item.js
@@ -452,34 +452,51 @@ export default class ContentItem extends Component {
 				}
 			}
 			
+			let AnyResults = false;
 			let ResultLines = [];
 			for ( let idx = 0; idx < Lines.length; idx++ ) {
 				let Line = Lines[idx];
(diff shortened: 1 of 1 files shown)


local-minimum:
Added handling of not connecting to server for final results too. Again, can't really test beyond `make`.

povrazor:
Sorry for the delay. Can you explain what it's supposed to do? Always a bit reluctant about untested code. ;)

local-minimum:
Hard to remember now, but I think that when there was some issue with the site last event all of the sudden the game-page said "You don't have the votes" even if you had the vote, when the issue was really "Site isn't responding at the moment". I think this is was trying to differentiate those two.

povrazor:
Riiiiiight. That does ring a bell.

sgstair:
This is probably no longer necessary, as the underlying bug is fixed, and nodes should always have scores.

PR #1023, opened 2017-05-26
fixed the fun, cleaned up the config sample

povrazor:


public-api/vx/fun.php
```diff
@@ -7,7 +7,7 @@
 json_Begin();
 
 // Bail if not the development environment
-if ( !defined(SH_PHP_DEBUG) && !SH_PHP_DEBUG ) {
+if ( !defined("SH_PHP_DEBUG") || !SH_PHP_DEBUG ) {
 	json_EmitFatalError_Forbidden("Sorry, this server isn't fun", $RESPONSE);
```
local-minimum: 😄

PR #1028, opened 2017-06-01
Sort medals

local-minimum:
This sorts the medals in the gold - silver - bronze order.
Can't test on local site but have tested that it can build and have tested code assuming that each array item is a string.


povrazor:
Okay. I'm going to merge this, but I'm actually going to disable it. At the moment I actually prefer the unsorted order, but I might eventually change my mind.

PR #1029, opened 2017-06-03
[Untested] Modified display of results

zwrawr:
Some small modifications to the results screen. 
Currently untested (haven't opened VM in a month, so it might be fun getting back on track with that)

* Changes
    > Graphics: **414** (2.778 average in 29 ratings)
    to something more like
    > Graphics: **414**th (2.778 average from 29 ratings)

### DIFF
--- a/src/com/content-item/item.js
+++ b/src/com/content-item/item.js
@@ -131,6 +131,16 @@ export default class ContentItem extends Component {
 		}
 	}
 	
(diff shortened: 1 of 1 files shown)


local-minimum:
It looks correct to me, it took me a turn of thinking before I understood the `k` was eliminating 11, 12, 13 on each 100.

PR #1040, opened 2017-07-12
the rest of the LD39 pre-event tweaks. sorry it's late

povrazor:


src/com/header-whatsup/whatsup.js
```diff
@@ -39,7 +39,14 @@ export default class HeaderWhatsup extends Component {
...
 			if ( featured.meta ) {
-				if ( featured.meta['event-theme'] ) {
+				if ( parseInt(featured.meta['theme-mode']) == 1 ) {
+					return (
+						<div class="header-base header-whatsup outside">
+							<span class="-title _font2">ON NOW:</span> {featured.name} <NavLink href={featured.path}><SVGIcon baseline small gap>suggestion</SVGIcon>Theme Suggestions</NavLink> are open!
+						</div>
+					);
+				}
+				else if ( featured.meta['event-theme'] ) {
```
local-minimum: This looks like it ought to be using `theme-mode` too.

povrazor: Not exactly (`theme-mode` only covers theme selection, not actual announcement), but in general the states of an event aren't easy to decode.

local-minimum: true, there should be an `event-mode` where `0` is the whole `theme-mode` cycle or something.

povrazor: Yeah. Running an event was pieced together one piece at a time as it was needed, so the overarching state machine wasn't planned out that well. ;)

Issue #1026, opened 2017-05-31
Calendar should respect local first day of week


local-minimum:
Perhaps the calendar should start with the day of the week that is standard for the country of the user. Or use the most common and ISO-standard, aka Monday.

povrazor:
I was under the impression it was Sunday.

povrazor:
Ah, lo and behold.

![funday](https://user-images.githubusercontent.com/1570313/27360233-e8dfc8b0-55ed-11e7-8581-b692a25452de.png)

Issue #847, opened 2017-05-03
Calendar Month Not Correct


ghost:
When you hover over the judging date on the calendar, it says June 17, 2017 instead of May 17, 2017.
I live in the Midwest United States if that makes a difference (CDT).
![2ae28b257cb1ffe470eb4592e4c33a2a](https://cloud.githubusercontent.com/assets/5943561/25647294/65df5a8a-2f86-11e7-9c6f-e98a12d74ab9.png)

zwrawr:
![image](https://cloud.githubusercontent.com/assets/18352787/25698033/30ee2fc4-30b5-11e7-8a88-4b39a7c2360d.png)

Happens for me too.  

I'm guessing it to do with the calendar not displaying multiple moths correctly.  
![image](https://cloud.githubusercontent.com/assets/18352787/25698087/63de1002-30b5-11e7-9ab9-e41fdf89c082.png)
![image](https://cloud.githubusercontent.com/assets/18352787/25698105/731b9292-30b5-11e7-8210-10f86a588516.png)

Seems to set the first cell to the current month, then increments the month.

povrazor:
This could be related to the end-of-the-year bug I experienced. I had to introduce a hack to temporarily change the highlighted day to the following month. But than stopped working once April 30th got pushed off the display.

I'm guessing what's happening is when the current day of the week is in the beginning of a month that isn't all 7 days, we forgot to subtract 1 from the month.

Issue #546, opened 2017-04-06
Calender: When the date changes, it should update


povrazor:
Calender should update automatically. Unilke the countdown clock, it only needs to update every day. But that said, if the site polled the server every our (while page is active) for schedule changes, that wouldn't be so bad.

Issue #500, opened 2016-12-18
Calender: Has 30 days in December, not 31


mikekasprzak:
Yay. This is getting silly.

Aurel300:
Isn't this fixed?

![screen shot 2017-04-16 at 18 15 46](https://cloud.githubusercontent.com/assets/1102162/25072841/c69b0ea4-22d0-11e7-873c-f4fba111c306.png)

povrazor:
It's not December. ;)

Aurel300:
It's not the 31st either. I set the Date to `new Date("2016-12-31")` in the above screenshot, tried a bunch of dates in December around that, also in 2017.

povrazor:
Oh, then I might have been seeing things. :wink:

Honestly LD37 was a bit of a blur. :tired_face:

PR #1051, opened 2017-07-20
Calendar fixes & minor other changes

sgstair:
Bugs:
* Fixes calendar bugs #500, #546, #847 
* Partial fix for #1026 (code now exists to display Monday-first, but no condition for enabling it is added)
* Fix infinite spinner on empty feed #748 

Side things:
* Add space under noob header (looks better)
* Add trophy icon to LD39 event in calendar (it was set on the wrong day)


povrazor:
Cool. That looks nice and clean. Merging...

PR #1056, opened 2017-07-21
Async/await support

Noojuno:
Could probably do with someone trying to convert some existing promise code to make 100% sure it's working.

I'll do at some point if nobody else wants to.


povrazor:
Okay, so you're not using the rollup nodent plugin?

PR #1061, opened 2017-07-25
Minor fixes

sgstair:
Fixed issue #748 again, now tested to work correctly with posts in the system
Suspected fix for all games showing up in the LD-specific games feed (missing parent/superparent method), but I can't easily test. Please confirm.
Added basic  website to automate adding an event in the test server. (sandbox/dev_events.php)

### DIFF
--- a/sandbox/dev_events.php
+++ b/sandbox/dev_events.php
@@ -43,12 +43,20 @@ function loadEvents()
 	$suggestname = "Testumdare " . ($eventcount+1);  
 }
 loadEvents();
-print_r($events);
+//print_r($events);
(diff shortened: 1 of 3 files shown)


povrazor:
The games definitely have parents (otherwise the website wouldn't work right). I will take a closer look at this in the morning.

sgstair:
Yes, the games have parents but the query does not currently use the parent method to restrict the feed to games with the right parent. So LD39 games shows all LD38 games at the moment.

povrazor:
Ah doh! Okay.

PR #1076, opened 2017-07-27
Improve submission links

Jezzamonn:
This is unfinished... I kinda got stuck figuring out the whole state updating thing in preact. I figured I'd make the pull request early so you can see my progress or explain how things work :)


Jezzamonn:
Alright, I may have got this working. Currently my VM is broken due to the new way the api is handled, so I haven't tested this at all, but if I revert to an older version I should be able to do that.

Also still to be done is how a game displays the links when it's not editing, that shouldn't be very hard though.

povrazor:
New way the API is handled?

Jezzamonn:
I briefly mentioned it on gitter, but basically no api calls were working due to cross domain stuff, so I couldn't test whether it worked.

Jezzamonn:
Ok, figured out what was going wrong.

Now I'm adding the server side stuff. These urls can now be saved, but they don't appear when the page is reloaded. 🤔

Jezzamonn:
Last thing left to do is make it so when you edit the links, the save button changed from "saved" to "save". Otherwise it looks like everything was automatically saved.

However, that button is part of ContentSimple (the edit thing used for posts) and it's state is managed by that ContentSimple's state, so I'm not sure if there's any way to update that.

I think The React Way would be to have that editing stuff dumb and not handle state, and then whenever you want an edit thing you need to add some parent thing that handles state.

Jezzamonn:
An alternative would be to update this stuff as it's edited, same as the other game stuff. Joining it with the save button can happen later, I guess

Jezzamonn:
Ok, should be good to go! Have a look at the final diff because I made some changes to some of the other text on the page too to make it feel more complete

src/com/content-item/item.js
```diff
@@ -66,7 +66,7 @@ export default class ContentItem extends Component {
 		});
 		for (let i = 0; i < 5; i ++) {
 			this.state.linkNames[i] = node.meta['link-0' + (i+1) + '-name'];
```
povrazor: // TODO: Make this support more than 9 links

src/com/content-item/item.js
```diff
@@ -66,7 +66,7 @@ export default class ContentItem extends Component {
...
-			this.state.linkUrls[i] = node.meta['link-0' + (i+1) + '-name'];
+			this.state.linkUrls[i] = node.meta['link-0' + (i+1) + '-url'];
```
povrazor: Don't use `-url`. Instead, have no suffix. Assume that `link-01` is a URL, and if no target (`B`) is set on the link, that `link-01-name` is how to set a name. When `B` is non-zero, the Tag node's `name` is used instead.

public-api/vx/node.php
```diff
@@ -28,6 +28,17 @@
 		'grade-06-out' => ['length' => 1],
 		'grade-07-out' => ['length' => 1],
 		'grade-08-out' => ['length' => 1],
+		
+		'link-01-name' => ['length' => 64],
+		'link-02-name' => ['length' => 64],
+		'link-03-name' => ['length' => 64],
+		'link-04-name' => ['length' => 64],
+		'link-05-name' => ['length' => 64],
+		'link-01-url' => ['length' => 128],
```
povrazor: Gosh, I wish we could limit URLs to 128 characters, but it'll have to be longer. I need to do a quick check first, but I'm guessing `2048` will cover us. Also it should be `link-01` without the `-url` (as the `-name` feature will become less and less needed as we populate our platform tag list). ;)

povrazor: Yeah, go with `2048`.

pschichtel: https://stackoverflow.com/questions/219569/best-database-field-type-for-a-url/219664#219664

povrazor: Internally it is `Text`, but this limit is protection to mitigate attacks.

povrazor: Actually no, lets go with `512`, and wait for someone to complain.

Jezzamonn:
FYI I deleted the showPrePub thing ("Hey folks! We're still finishing the data fields below ..."). Don't think it's needed anymore.

povrazor:
Sounds good.

Jezzamonn:
Yay!

PR #1147, opened 2017-08-01
potential fix for ALL performance problems... damn

povrazor:


local-minimum:
This change makes me wonder, is it worth going through the code and checking if other things could look for cached data?

povrazor:
This is a lot of what I've been doing already. Care needs to be take though, as caching and using the same account on 2 computers is going to break

PR #1177, opened 2017-08-21
changed end time and filter

povrazor:


src/com/view-content/content.js
```diff
@@ -591,7 +591,7 @@ export default class ViewContent extends Component {
 			else if ( extra && extra.length && extra[0] == 'games' ) {
 				let DefaultSubSubFilter = 'featured';
 				let DefaultSubFilter = 'all';
-				let DefaultFilter = 'smart';
+				let DefaultFilter = 'danger';//'smart';
```
local-minimum: If this component could know about `timeUntilVotingEnds` is could easily be fixed so that `'danger'` is only used last 48h and `'smart'` is used all other time.

PR #1214, opened 2017-09-12
Cleanup of component-item

local-minimum:
This PR removes voting related things from the component-item and places them in smaller components of their own. I've tried around some on events in various stages and think that I've captured all the different ways they are used.

One potential concern is how/if the `VoteOrResults` in `com/vote-modal/modal.js` reacts to user clicking forwards and backwards in the browser between game pages.


povrazor:
These should all be under the same folder since they're only used in one place (item/game pages), and "vote" is not a highest tier item like content and views. (sorry I'm heading out the door but I'll be back in an hour and can elaborate)

local-minimum:
I kind of guess that they perhaps should be moved, I wasn't sure where though so just let me know and I move them accordingly.

povrazor:
Well they are things that came out of content-item, so they should be with
content-item.

local-minimum:
Alright now I've moved the files.

src/com/content-item/vote-voting.js
```diff
@@ -0,0 +1,117 @@
...
+				{ShowRatingSubText}
+				<div class="-items">{VoteLines}</div>
+				<div class="-footer">Ratings are saved automatically when you click. When they change, they're saved.</div>
+			</ContentCommonBody>
+		);
+
+	}
+	
+	
+}
```
povrazor: Add newline (linter).

src/com/content-item/item.js
```diff
@@ -64,19 +65,9 @@ export default class ContentItem extends Component {
...
-		.then(r => {
-			if ( r.grade ) {
-				this.setState({'grade': r.grade});
-			}
-			else {
-				this.setState({'grade': []});
-			}
-		})
 		.catch(err => {
 			this.setState({ 'error': err });
```
povrazor: Prefer `({'error': err})` (note the lack of prefix and suffix spaces within the `{}`'s). A very minor change, but I'll have an explanation later.

src/com/content-item/item.js
```diff
@@ -726,13 +381,16 @@ export default class ContentItem extends Component {
...
 		}
-
-		var ShowLinkView = null;
-		if ( true ) {
+		// Where you can enter your game links
+		let ShowLinkEntry = null;
+		let ShowLinkView = null;
+		if ( editing ) {
+			ShowLinkEntry = this.makeLinks(true /* editing */);
+		} else {
```
povrazor: Prefer
```
{
else {
```

src/com/content-item/vote-current.js
```diff
@@ -0,0 +1,55 @@
...
+			<ContentCommonBody class="-rating">
+				<div class="-header">Total Ratings</div>
+				<div class="-subtext">Votes on your game so far</div>
+				<div class="-items">{VoteLines}</div>
+				<div class="-footer">To get a score at the end, you need about <strong>20</strong> ratings in a category. To get ratings: play, rate, and leave feedback on games. Every game you rate and leave quality feedback on scores you <strong>Coolness</strong> points. Having a high "Coolness" prioritizes your game.</div>
+			</ContentCommonBody>
+		);
+	}
+	
+}
```
povrazor: Newline (linter).

src/com/content-item/vote-metrics.js
```diff
@@ -0,0 +1,99 @@
...
+				<div class="-subtext">Advanced data on this game</div>
+				<div class="-items">
+					{SimpleLines}
+					{AdvancedLines}
+				</div>
+				<div class="-footer">Metrics update rougly every <strong>15 minutes</strong>. If they don't exactly match other data (i.e. ratings), this is because they haven't updated yet. Metrics with a <strong>*</strong> are dynamic, and change based on a variety of factors. It is normal for these numbers to go up and down.</div>
+			</ContentCommonBody>
+		);
+	}
+}
```
povrazor: Newline (linter).

src/com/content-item/vote-modal.js
```diff
@@ -0,0 +1,64 @@
...
+import VoteVoting						from 'com/content-item/vote-voting';
+
+
+export default class VoteOrResults extends Component {
+	
+	constructor( props ) {
+		super(props);
+	}
+	
+	render(props, state) {
```
povrazor: Prefer `render( props, state ) {`, noticing the added spaces before the arguments. 

This is specifically in this case, because it defines a function. It's a subtle visual cue that this use of `blah( ... )` is defining a function instead of calling it.

Calling functions should have no prefix/suffix spaces (i.e. `render(props, state);`). 

`if` and `for` blocks should be treated similar to function definitions, but are sort of their own thing, as there should always be a space after a keyword (to help differentiate it from a function call). 

i.e. `if ( true ) {` or `for ( let idx = 0; idx < 10; idx++ ) {`.

src/com/content-item/vote-optout.js
```diff
@@ -0,0 +1,105 @@
...
+					if ( r && r.changed ) {
+						Node.meta[Name] = Data[Name];
+						this.props.optOutCallback();
+					}
+					return r;
+				});			
+		}
+	}
+	
+	render(props, state) {
```
povrazor: `render( props, state ) {` 

(see other comments on defining functions)

src/com/content-item/vote-results.js
```diff
@@ -0,0 +1,79 @@
...
+	constructor( props ) {
+		super(props);
+	}
+
+	hasFinalResults() {
+		const nodeComponent = this.props.nodeComponent;
+		return !parseInt(node_CanGrade(nodeComponent)) && node_isEventFinished(nodeComponent);
+	}
+	
+	positionSuffix(position) {
```
povrazor: `positionSuffix( position ) {`

src/com/content-item/vote-results.js
```diff
@@ -0,0 +1,79 @@
...
+	    let j = position % 10;
+		let k = position % 100;
+	
+	    if (j == 1 && k != 11) return "st";
+	    if (j == 2 && k != 12) return "nd";
+	    if (j == 3 && k != 13) return "rd";
+	    return "th";
+	}
+	
+	render(props, state) {
```
povrazor: `render( props, state ) {`

src/com/content-item/vote-voting.js
povrazor: newline (linter)

local-minimum:
Since this contains drop downs and I have other PR changing them some, I think it is better to redo cleaning from start one that is in. So I close this PR.

PR #1215, opened 2017-09-14
Added tootips

local-minimum:
## Features

```jsx
<ToolTip PopUpContent={<something />}  />
```

There are a bunch of optional properties to it:
* `ToolTipButtonContent` overrides the default SVG info-icon
* `maxWidth` (default 0.5) sets the maximum width as fraction of the width of the viewport
* `horizontalPadding` (default 6px) number of pixels padding to the edge of the viewport for the popup
* `hideDelay` (default 500ms) a time after mouse exits until toottip is hidden (this helps being able to move mouse from the icon that triggers the tooltip to the popup box. Which is useful if it contains buttons or such).
* `touchNoDoubleClick` (0.1s) a delay for when it isn't possible to hide the tooltip with a click after it has been shown. To counter touch causing flicker show-hide effects.

There's also a subclass for easier use if only text is intended (can be paragraphs and such too ofc) to be shown. This one will resize its width to minimize the number of lines in the tooltip while respecting the conditions set by above parameters.

```jsx
<ToolTipText Text={'My text is fine'} />
```

## Known issues
It is a bit weird that so much styling is done through the JS, but I found no way to make the content of the tooltip stretch out in a reliable way.

Therefore, the `ToolTipText` can sometimes give a bit strange padding on the right side.

I could also not reliably make the tooltip show on top of the right-side bar, so if the tooltip is triggered quite far to the right but isn't too wide to force it to shift alignment, it could be partially hidden.

### DIFF
--- a/src/com/tooltip/tooltip-text.js
+++ b/src/com/tooltip/tooltip-text.js
@@ -1,0 +1,31 @@
+import { h } 				from 'preact/preact';
+
+import ToolTip						from 'com/tooltip/tooltip';
+
+
+export default class ToolTipText extends ToolTip {
+	constructor( props ) {
+		if (!props.lineHeight) {
+			props.lineHeight = 18;
+		}
+		
+		this.textSpan = null;
+		props.PopUpContent = <div class='-tooltip-text' ref={(span) => this.textSpan = span}>{props.Text}</div>;
+		super(props);
+	}
+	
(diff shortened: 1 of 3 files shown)


povrazor:
I'll merge this, but this should be renamed to something more consistent (i.e. all com things have are formatted `type-name`, and if they're not they need to be fixed).

PR #1216, opened 2017-09-19
Simulator cache fixes

sgstair:
* Change `vx/node/meta` to allow Administrators to edit some event meta
* Administrator is now defined as a user with meta `"can-create" = "event"`, as well as the existing` $USER['private']['admin']=1` (which I'm not sure if this is produced anywhere)
* Improved strictness of meta/link type information, to prevent any mistakes from causing permission leaks.
* Modify event simulator to use API to set some meta (as this results in the node being flushed from cache, and solves the caching problems the simulator was having, when running faster using opcache.
* Add a website to sandbox (`dev_weblogs.php`) for getting the apache error/access logs.


povrazor:
I'm still working my way through, but I don't think I like the idea that a defining factor of an admin is that they can create events (i.e. inside `userAuth_IsAdmin`). Right now yes it is restricted, but eventually others will be able to create events, just not under the Ludum Dare folder.

povrazor:
Yeah, that's the main problem I'm seeing.

If `$USER['private']` is not available, then that's probably a bug.

What's supposed to happen is you have `['meta']` for the public metadata, `['protected']` for the metadata that only you can see, and `['private']` for only things the server can see.

povrazor:
privvy

Ah. So that's why it didn't work (the fetch is commented out).

povrazor:
FYI I'm working on a patch that will make `shared`, `private`, and `server` user data available as part of `$USER`. Should be ready in a few hours.

PR #1217, opened 2017-09-19
Notification API

sgstair:
(Includes commits from earlier pull request, I may have to rework this PR if that one is committed first, but I don't know how cautious you will be about accepting this PR :) )

**Quick Summary**
* Added /vx/notification API surface with 5 distinct calls. Count+feed for all+unread, and a mark read API
* Mark read updates a read cursor (stored in user table) which is the ID of the highest read ID, any newer IDs are considered unread.
* Posting a note will produce notifications for the node author, and all other note-posters in the thread
* Posting an item, or a post, will produce notifications for any users who have starred the author.
* The new API functionality was added to LD event simulator and tested relatively extensively.
* As there are DB changes, you will need to update tables (www/src/shrub/tools/table-create - or rebuild with vagrant delete / vagrant up)

* Note this change also includes changes in [PR#1216](https://github.com/ludumdare/ludumdare/pull/1216), which it turns out this code has a dependency on due to simulator changes. That PR includes code to allow an admin user to set event metadata.

**API Details**
* GET `/vx/notification/unread/count` and `/vx/notification/all/count` return the total count of notifications.
```
{
    "status": 200,
    "caller_id": 213,
    "count": 33
}
```
* GET `/vx/notification/unread/feed` and `/vx/notification/all/feed` allow for returning sections of the notification list. GET Parameters are `offset=` and `limit=`. Limit may be up to 100
Feed items are returned with the most recent being index 0. 
Note that this call also returns a `count` field with the total count of notifications (either all or unread depending on the call)
The current `type` values are "note", "post" and "item", where "post" and "item" are just types of nodes. The type is an 8-character field in the DB and we can add other identifiers in the future for new kinds of notifications.
The reason for having a date field in the notification record, is mainly for these other kinds of notifications in the future. As with the existing types, the date can be inferred from the node/note.
```
{
    "status": 200,
    "caller_id": 213,
    "offset": 0,
    "limit": 20,
    "count": 33,
    "feed": [
        {
            "id": 5929,
            "node": 359,
            "note": 1399,
            "type": "note",
            "created": "2017-09-19 08:38:21"
        },
        {
            "id": 5907,
            "node": 324,
            "note": 1391,
            "type": "note",
            "created": "2017-09-19 08:38:20"
        },
...
```
* POST `/vx/notification/markread` with parameter `max_read=<id>` tells the system that the user has read up to the id specified, and only newer notifications should be included in "unread".
This ID is range checked and the call will fail if you try to set the cursor below the current cursor (can't mark unread), or if you set a value higher than the current highest notification for the user (can't mask future notifications)


local-minimum:
# First thing

> Posting`

 a note will produce notifications for the node author, and all other note-posters in the thread

Two questions:

a) Will it notify all authors of the node (i.e. all collaborators on a game)
b) Won't a comment on a node generate a bit too many notifications if the node is a popular node (i.e. a managore game?) if everyone who has a note on it will get a notification? Or do you see it as a UI thing to filter this depending on user settings or simply filter note-to-note notifications on at-usage?

# Second thing

Is the API entirely public or is it checking auth to get some of the info.

sgstair:
a) It will not currently notify all the authors for comments on a game node. Good catch, I recall we talked about this but forgot to implement it.
b) It will generate a lot of events in some cases, yeah. The way I roughly see it is some kind of icon with a number in your top bar, and clicking there you can browse through all the events, but yes, that is likely to include a lot of events for some of the more popular threads. I'm not sure if this is actually a problem. 

Maybe we should also add a mechanism to coalesce multiple note notifications, that could be either UI or API. Also we might want to consider adding a table to allow users to disable notifications for a noisy node.

Ah, I didn't mention, but the entire notification API surface is user-dependent. It's meaningless without a user as it's retrieving per-user data.

local-minimum:
About user dependent, I was hoping as much, the question was more about the ability to track and monitor peoples activity on the site, and while the information is publicly available, I think it is good that this tool is not.

I will probably find the answer to this myself, but one question is if you can generate notifications for yourself?

local-minimum:
First attempt I get: `{"status":500,"response":"Internal Server Error","message":"See Log [59c0ff5f476a5]"}{"status":500,"caller_id":9,"count":null}` but I'm guessing this is missing a database?

local-minimum:
How about if someone voted on your game and not commented, should that trigger notification.
How about if you participated in any step of an event and the event-status updated?

(Not saying these need or even should be included now, but just keeping here for reference)

sgstair:
When you post a note, your user ID is filtered out of the list of user IDs to notify, so no, you can't notify yourself. If you were to manually star yourself through the API (not possible in the UI), then you would get notifications for publishing your posts and items. I didn't filter this one out.

Voting does not currently push a notification, I don't think it should personally. Also if we want to notify on event state change, I think a different mechanism would be better.

sgstair:
I realized there is a bug in the notifications, notes will double notify if the node author also posts in the thread. I'll resolve this and update the PR this evening. I'll also probably notify all linked authors rather than just the node author (allow all game authors to get notifications for comments on the game)

sgstair:
Added & Tested fix.

local-minimum:
Notifying all authors seems like a reasonable thing.

One column that may be nice to have from start in the notification had what user id did the action that caused the notification.

Also, is there a difference between an update to a node or note compared to initial post in the notification or if not visible on the node /note

sgstair:
I've been trying to keep the notification record as minimal as possible because it grows at about 10x the speed of say "note" in the simulator, which is probably moderately close to how the site will operate. Ultimately the information about what user caused the change is / will be available via other means - but just assuming the author in the node/note is responsible is good enough for now.

sgstair:
Oh, I forgot to mention that, updates do not cause notifications, only new posts. I may reconsider this if there are good arguments for changing it.

local-minimum:
@sgstair my thinking is that an update to a game with several authors should trigger notifications to the other authors of the game/node. I don't think it should trigger notifications to everyone who has interacted with the game/node nor everyone who has followed any of the authors.

sgstair:
I'm not sure I see a compelling argument for adding notifications in that scenario. Also I think a lot of node modifications are trivial and bunched up, like correcting minor typos after making an edit. If there's a major update they would probably post a comment, and that would lead to notifications anyway

sgstair:
Another thing remaining to do here is to handle @ mentions and produce notifications for them. I'm thinking a bit over how the code for this will work.

povrazor:
On post, we need to merge the data in to a blob to insert in to the search
blobs table. My plan then was to check for @names in the blob, and add a
notification instance pointing back at the post/comment. This insert only
happens on publish though (i.e. while published date is zero, and becomes
non zero)

povrazor:
Oh, and case it isn't clear: a search blobs is the appended title, body, and a bunch of other content like metadata link information (that can come later). Not sure about comments. The point of the appending is so the search is only 1 operation per node, rather than several (titles, bodies, metadata, etc). MySQL/MariaDB have a full-text index type, which in theory will give us our Google magic. 

It may make sense to full-text index the comment bodies themselves, but restrictions will need to be placed on scope for search (i.e. no searching of other users private messages).

Another thought was to actually strip certain data out of the text before storing in the search blobs, but because our source text is all markdown, it might not be too polluted.

sgstair:
Yeah, The way I was thinking of adding the notifications for @ mentions was to scan the body on post/publish, and old/new on edit (when published), and identify new @ mentions to post notifications for. I think this will be necessary for it to behave as people will expect. I also expect to scan the title text for post/game items. It is probably best being independent of the search system, imo.

povrazor:
I just mean there is a list of actions that happen on publish. Right now it's add a version, then update the node. To be added is updating search blobs, which is basically appending two+ strings together (name and body). That appended string may be useful.

local-minimum:
For now since we actually trigger notifications for most cases where mentions are prone to happen, I made the client side detect mentions in comments/notes.

local-minimum:
But only in the body

povrazor:
The changes to the `sh_user` table need to be removed. `sh_user` is a high risk table that contains password hashes and e-mail addresses. Use of this table needs to be restricted for security.

It also contains many dead users (un-activated accounts). Not a big deal in this context, but IDs of the user table don't directly match actual active users.

sgstair:
Understood (and we talked about it on gitter, I'll revise this soon) - Just posting here to note that the added access to sh_user is based on the node field, not id - so the users do match correctly.

povrazor:
I more meant that every account, even ones that aren't activated, would have that data associated with them. So it's basically wasted space.

Whatever. All I'm trying to say is it's not a good place for data, even though it's called `user`. :wink:

PR #1219, opened 2017-09-20
Support React Dev Tools when site invoked with ?debug

sgstair:
Title says it all.

### DIFF
--- a/src/external/preact-devtools/devtools.js
+++ b/src/external/preact-devtools/devtools.js
@@ -1,0 +1,391 @@
+/* global __REACT_DEVTOOLS_GLOBAL_HOOK__ */
+
+import { options } from '../preact/preact';
+
+// Internal helpers from preact
+import { ATTR_KEY } from '../preact/constants';
+
+/**
+ * Return a ReactElement-compatible object for the current state of a preact
+ * component.
+ */
+function createReactElement(component) {
+	return {
+		type: component.constructor,
+		key: component.key,
+		ref: null, // Unsupported
+		props: component.props
+	};
+}
+
+/**
+ * Create a ReactDOMComponent-compatible object for a given DOM node rendered
+ * by preact.
+ *
+ * This implements the subset of the ReactDOMComponent interface that
+ * React DevTools requires in order to display DOM nodes in the inspector with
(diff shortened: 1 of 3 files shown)


povrazor:
Cool. 

The `import` in main-ld.js should still include the code, it just doesn't execute it. Some of this checking may have to move to the makefile instead to outright omit it. 

(The way rollup works is everything imported is included, everything not imported is ignored).

sgstair:
Yeah, it's not much code though so it might be ok. I noticed it would be a significant amount of dev effort to actually exclude it from the main code package.

povrazor:
I am fine merging it in, especially given the benefits it'll have. But just FYI this is going to be one of those things that does bug me (and it's probably just me).

Like a general goal I have for this project is that the minified+gzipped size of the js, css, and SVGs, I want that under (or as close to) 100k as possible. There's a bunch of dead js and css code still lingering around, plus some IE and old Webkit compatibility stuff that can be isolated, and fetched only when running on legacy browsers. Maybe it's a futile task, but the smaller the binary, the faster our startup time is.

sgstair:
I agree on keeping the JS size low, I didn't check exactly what the impact was but I don't expect it was much as the added source is only 13k unminified, and it looks to be >50% comments. I guess that's a little bigger than I thought it was.

Two other things on size:
* I noticed that even in the minified js there are quite a lot of names (generally local functions) that don't get minified, and I wonder if there are some settings to be even more aggressive with renaming.
* I grudgingly admit that force push is useful, but it's kinda dumb. If there were a way to push the file times of those files rather than just the full file, that would be the best of both worlds. You'd take the round trip time hit on the first request, but on subsequent requests you'd know whether your cache is up to date instead of pinging the file with a HEAD request, and you'd avoid sending the superfluous large js+css+svg file contents every time the page is loaded.

sgstair:
ah, regarding my complaints about http/2 push they are already working on fixing this issue. Just not finished yet. https://datatracker.ietf.org/doc/draft-ietf-httpbis-cache-digest/

povrazor:
Yes minification can be improved by munging symbols. I'll add an issue for that.

PR #1222, opened 2017-09-22
Notification ui

local-minimum:
## Before merging this...

I based it not on the master but on an early version @sgstair API for notifications. So I'm not sure if the UI files should be pulled in independently to ensure newest version of the API is actually used. Also this PR should not really be merged without the API.

## Features

The notification bubble displays the number of unread notifications. Currently this is set to poll the API once a minute.

Clicking on the icon will show a drop-down of the 8 oldest unread notifications (or 8 most newest notifications if there are no unread). Clicking on any item will direct you to the node that triggered it. So for notifications coming from notes, you have to scroll to find them.

Once the notification drop-down has been loaded, those notifications are marked server-side as seen/read. For this reason, the drop-down stays expanded after clicking one of the options as there is currently no way of getting second chances for viewing notifications through the drop-down. The unread counter is immediately updated.

Clicking the bubble-icon will collapse the drop-down. And if you have more remaining notifications, clicking it again will retrieve the next batch. This is a bit annoying, and perhaps with using the drop-down we can get more than 8 without it being too much.

## Known issues

Currently a link to a `/home/notifications` is added to the bottom of the notifications. This is a 404 page, and that code should perhaps be commented out.

Because I don't know css/less well enough, I didn't find a smart way of making the drop-down wide enough that it felt good reading the comments while at the same time ensuring it can be viewed on all screens. For now the `.less` specify `480px`, this should probably be fixed somehow to ensure the left side never goes further than say 20px from left edge of the screen. I could add code similar to the code in the tooltip for collision detection if no better solution exists and if a better solution is needed.

## Comments

I probably didn't place the new code in the proper directories.

I needed to extend the flexibility of the drop-down a bit to hide the selected item area/functionality.


povrazor:
Heh, okay this is going to be a rough merge. :smile:

One step at a time.

local-minimum:
If you just grab the commits that are mine, it should work better or just pick the files that are js/less  (client side)

PR #1236, opened 2017-10-01
Relocate max_read notification to server private metadata

sgstair:
Also add max_read field to notification feeds.
Note the table create change leaves the field present but hanging in test environments, it shouldn't cause a problem. May require VM rebuild if we update that table further.


public-api/vx/notification.php
```diff
@@ -53,6 +52,7 @@
 							$RESPONSE['limit'] = 100;
 					}
 					
+					$RESPONSE['max_read'] = notification_GetLastReadNotification($user_id);
```
local-minimum: Does this add `max_read` to the response? Fo which calls in that case?

sgstair: Yes, it's added to all /vx/notification/*/feed calls.

PR #1247, opened 2017-10-05
Anonymous comments

sgstair:
This is the backend work to enable anonymous comments.  
Following are the major changes:
* Add `flags` to note and note version tables. Define two flags (hidden and anonymous)
* Add `allow-anonymous-comments` metadata that can be applied to an 'item' node - The expected value for "enabled" is `1`
* Add `anonymous` optional parameter to `/note/add`, posting anonymously will only succeed if `allow-anonymous-comments` is present and resolves to a nonzero / true value.
* `flags` as well as boolean convenience variables `anonymous` and `hidden` are returned to the client with notes
* Minor site JS changes to not get stuck when anonymous comments are returned, and to render them appropriately.
* Event simulator now marks some games with `allow-anonymous-comments`, and sometimes comments anonymously on them.

Unrelated to anonymous comments:
* Fixed bug causing 500 server error on `notification/markread`
* Add `note/getnote` for fetching one or more notes by ID (Mainly for notifications). It's expected that this will be renamed in the future with a reorganization.
* Add event simulator concept of 'reaction' - Newly posted posts may cause a collection of users to interact with the new node - this will make the simulated dataset look more like actual data with many consecutive notifications on the same thread (to assist with development of comment merging in notifications / etc)

### DIFF
--- a/public-api/vx/note.php
+++ b/public-api/vx/note.php
@@ -80,9 +80,9 @@ switch ( $action ) {
 
 
 			if ( isset($_POST['anonymous']) && $_POST['anonymous'] ) {
-				// Only allow anonymous comments to be posted if the node has "allow_anonymous_comments" meta set.
+				// Only allow anonymous comments to be posted if the node has "allow-anonymous-comments" meta set.
 				
-				$meta = nodeMeta_GetByKeyNode('allow_anonymous_comments',$node_id);
+				$meta = nodeMeta_GetByKeyNode('allow-anonymous-comments',$node_id);
 				if ( count($meta) == 0 || !$meta[0]['value'] ) {
 					json_EmitFatalError_BadRequest("Cannot post anonymous comment: Node doesn't allow anonymous comments.", $RESPONSE);
 				}
--- a/public-api/vx/note.php
+++ b/public-api/vx/note.php
@@ -32,9 +32,23 @@ switch ( $action ) {
 			json_EmitFatalError_BadRequest(null, $RESPONSE);
 		}
 
-		$RESPONSE['note'] = noteComplete_GetByNode($ids[0]);
+		$RESPONSE['note'] = noteFlags_Filter(noteComplete_GetByNode($ids[0]), userAuth_GetID());
 
-		break; //case 'get': //note/get
+		break; //case 'get': //note/get		
+	case 'getnote': //note/getnote
+		// Temporary "Get note by note ID" call. To be renamed when the renaming time comes.
+		json_ValidateHTTPMethod('GET');
+
+		$ids = explode('+', json_ArgGet(0));
+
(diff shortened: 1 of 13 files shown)


local-minimum:
What is the meaning of `hidden` compared to `anonymous`?

public-api/vx/note.php
```diff
@@ -32,9 +32,23 @@
...
+		json_ValidateHTTPMethod('GET');
+
+		$ids = explode('+', json_ArgGet(0));
+
+		// Limit number of notes
+		if ( count($ids) > 250 ) {
+			json_EmitFatalError_BadRequest("Too many notes", $RESPONSE);
+		}
+		
+		$RESPONSE['notes'] = noteFlags_Filter(noteComplete_Get($ids), userAuth_GetID());
```
local-minimum: I generally prefer plural for things that are lists, but I think the site generally uses `note` even when the returned items are lists with potentially many notes

public-api/vx/note.php
```diff
@@ -58,16 +73,30 @@
 			if ( isset($_POST['tag']) )
 				$version_tag = coreSlugify_Name($_POST['tag']);
 
+			$parent = intval($_POST['parent']);
+
 			// Load Node
 			$node = node_GetById($node_id);
 
+
+			if ( isset($_POST['anonymous']) && $_POST['anonymous'] ) {
```
local-minimum: I'm wondering if the note post should not be refused if it is set to be anon when the node is not allowing it. Looks to me like it would be posted non-anon then. This could happen if the node author changes the setting while you are composing your comment

local-minimum: sorry now I see that's exactly how it works

sgstair:
`hidden` is the first stage of future features around automatic moderation on flagging - The way I see it, moderated comments will still be available, but just hidden by default so you have to interact with the comment to show them. Probably there could also be stronger types of flagging that remove them (`deleted` or other things in the future)

povrazor:
It should hide posts. TBD where we'll use it, but it may make sense to hide a post. 

Oh also, that reminds me. We should add a field for comments that are not kid friendly. It doesn't need to be used yet, this may be something we later introduce for strict parents/teachers.

public-api/vx/note.php
sgstair: That's true, the one other location that returns notes does return with `note` :P I'm not too worried, this path is going to be reworked.

sgstair:
Also to answer the last unanswered question which I glossed over. Notifications don't contain the author of the change that caused the notification - And the only way to get the note object is through those two pathways which I added the filter function to. If we return notes in the future, they should also be filtered (it isn't done by default, because we might want to use that anonymous author data server side or in tools)

PR #1248, opened 2017-10-05
Notification feed copy

local-minimum:
This implements a `/home/notifications` feed and restructures existing notification code to not duplicate too much.

There's a bit of routing for the uri to show the feed page.
There's a new `$Note.Fetch` in shrub that gets the individual notes needed.

There are some clear areas of further improvements:

1. Color/styling of the notification feed needs more work. Especially when there are new notifications.
2. The drop-down should be collapsed when clicking to see the feed.
3. The number of new notifications in the notification-bar should be synced with pressing mark all read in the notification feed.
4. The notification button in the bar should not expand the drop-down if user is viewing the notification feed, but rather reload the feed contents / add new notifications if existing.
5. The user may get confused if the notification count says 4 but there are only two notifications due to 3 of them being collapsed into a single `3 users also commented on blablabla`. This could be helped with some number in a circle showing the number of notifications that went into that notification.


povrazor:
Okay, so like we mentioned in #1249, here's a little post merge report for you.

For the most part, since this is new code, I let it in mostly as-is. Just a couple style related issues I thought I'd address more formally.

* For simplicity I open a new merge request (#1251) that contained everything found in this one. GitHub lets me commit changes made right from the GitHub UI to PR's, but I'm not sure how to do it from the command line (yet). (It might be possible but the document I was reading wasn't very clear)
* For readability in CSS, I added `&` at the beginning of a couple lines that were `> a` blocks. Yes functionally they work the same, but CSS files (LESS files) get large, so it's more just a nicety that when you see an `&`, you know its the child of something.
* I introduced a bunch of style tweaks in #1249, things I'd mostly been following, but mistakes happen. For the most part they are changes to how whitespace is used. The code should be the same as it was before, but since changes to whitespace can trigger unnecessary merges in files, we can simplify merges by forcing the style to be more stricter. As an example, `src/com/input-dropdown/dropdown.js` didn't contain any changes at all, just some added whitespace. But if we're every looking backward in time tracking an issue, we've potentially wasted a moment looking at this file that wasn't really a part of the change. In hindsight, what I should have done in #1251 was 

Style things of note:
* all files should end with a blank line (i.e. a newline). Yes tools don't typically have a problem reading them, but it's good form to end any text-based file with a blank line. Tools like GitHub's merge tool output a little :x: on the last line of files, so they tend to jump out at me when I see that.
* Lines should not end with whitespace. It can be very difficult to even notice when this happens, so the linter will now warn you about it. This might become one of the more annoying changes, but end-of-the-day it will mean less false positives for unchanged files.
* indentation (tabs vs spaces) should not be mixed (tabs or spaces only). I would prefer indentation be all-tabs (4 characters wide), but the filter isn't that aggressive yet. Now this isn't python, so indentation wont break anything, but like lines ending with whitespace, lines beginning with mixed whitespace are candidates for false positives when merging.

That's about all that comes to mind at the moment. I haven't touched anything regarding function, just trying to improve some of the practices when I see them.

local-minimum:
For new code the linter is really insisting on that there are no trailing spaces. That's good.
I'm not sure exactly what you mean by the indentation, 4 spaces or something else 😢 ?
Otherwise, I'm agreeing. 
And I appreciate the detail in your comment, but it is also fine to focus on one or two things that you would like to see fixed.

povrazor:
If you insert a "TAB SPACE TAB" you'll see what I mean.

src/com/content-notifications/notifications.js
```diff
@@ -0,0 +1,92 @@
...
+		});
+				
+		const ShowGetMore = hasMore ? (<ContentMore onclick={this.fetchMore} />) : null;
+			
+		const ShowSetAllRead = hasUnread ? (
+			<ButtonBase 
+				class="-button -light focusable"
+				id="button-mark-read"
+				onclick={(e) => {this.markReadHighest();}}>				
+				Mark all commentes as read
```
sgstair: 'comments' rather than 'commentes'

local-minimum: good catch! @povrazor should I add it to one of my PR or you fix?

PR #1249, opened 2017-10-05
New stricter ESLint rules. sorry ;)

povrazor:


local-minimum:
This is great, I would like I said before also appreciate code review on github to be part of the process.

povrazor:
Do you have a particular example of what you're looking for in a review?

local-minimum:
I hade used `className` rather than `class`.  As an example and I might very well not have looked at you fix.

But also in general, though it requires more of the person doing the pr, it is also a way to teach each other and learn. even if frustrating at times. 

I am also thinking that if review was more of a thing it would spread the responsibly a bit even if you have the final check/say so.

povrazor:
Okay, I think I get it. So it's more a request for a report after talking about what and why I changed things, and highlighting things to watch out for.

Is it fine that I do it right in the closed/merged PR? Like this one, because we're both talking, we're both getting replies sent to us. Let me go give an example...

local-minimum:
I get all notifications for all things on LD, so for me it is fine. In general it should notify the person who did the PR, so yes. I think that is fine. And to be clear, it is just a suggestion

povrazor:
It's a good suggestion. :+1:

PR #1253, opened 2017-10-07
Fixing specific columns and large game

local-minimum:
This introduces a hammer drop-down where number of columns to the games-grid can be selected.

At the moment the `<ContentsGames>` has a default layout of 3 columns for whatever platform in its state.
When there's a parameter for if on mobile, the `defaultLayout` should probably be moved to `props` or `props` should allow for a flag of default view-type to trigger different defaultLayout settings.

The hammer-button that triggers the layout should probably in the near future be placed inside the filter/search bar that most instances will have. That is for a future PR.

The PR also ensures that the number of boxes/cards produces a complete final row with invisible placeholders should need be. This fixes the bug of sometimes getting extra large boxes/cards on the final row.

### DIFF
--- a/src/com/content-box/box.js
+++ b/src/com/content-box/box.js
@@ -153,7 +153,8 @@ export default class ContentBox extends Component {
 					</div>
 				</ButtonLink>
 			);
-		} else if (props.placeHolder) {
+		}
+		else if (props.placeHolder) {
 			return <div class={cN(Class, props.class, '-place-holder')} />;
 		}
 		else {
--- a/src/com/content-box/box.js
+++ b/src/com/content-box/box.js
@@ -50,11 +50,7 @@ export default class ContentBox extends Component {
 
 	render( props, state ) {
 		props = Object.assign({}, props);
-
-		var node = props.node;
-		var user = props.user;
-		var path = props.path;
-		var extra = props.extra;
+		const {node, user, path, extra} = props;
 
 		if ( node /* && state.authors */ ) {
 			var Class = ["content-box"];
@@ -133,7 +129,7 @@ export default class ContentBox extends Component {
 //				ShowTrophies.sort(function(a, b) {
 //					sortOrder = ['-first', '-second', '-third'];
 //					regexPattern = /class=\"([^\"]+)/;
-//					return sortOrder.indexOf(regexPattern.exec(a)[1]) - sortOrder.indexOf(regexPattern.exec(b)[1]); 
+//					return sortOrder.indexOf(regexPattern.exec(a)[1]) - sortOrder.indexOf(regexPattern.exec(b)[1]);
 //				});
 			}
 
@@ -157,6 +153,8 @@ export default class ContentBox extends Component {
 					</div>
 				</ButtonLink>
 			);
+		} else if (props.placeHolder) {
+			return <div class={cN(Class, props.class, '-place-holder')} />;
 		}
 		else {
 			return <ContentLoading />;
--- a/src/com/content-games/games.js
+++ b/src/com/content-games/games.js
@@ -1,5 +1,4 @@
 import { h, Component } 				from 'preact/preact';
-import NavSpinner						from 'com/nav-spinner/spinner';
 
 import ContentLoading					from 'com/content-loading/loading';
 import ContentError						from 'com/content-error/error';
@@ -12,8 +11,6 @@ import ContentCommonBodyTitle			from 'com/content-common/common-body-title';
 
 import GridSelector						from './grid-selector';
 
-//import ContentPost						from 'com/content-post/post';
-//import ContentUser						from 'com/content-user/user';
 import ContentMore						from 'com/content-more/more';
 
 import $Node							from '../../shrub/js/node/node';
@@ -24,13 +21,13 @@ export default class ContentGames extends Component {
 		super(props);
 
 		this.state = {
-			feed: [],
-			hash: {},
-			offset: 12-5, //10-5
-			added: null,
-			loaded: false,
-			defaultLayout: 3,
-			layout: 3,
+			'feed': [],
+			'hash': {},
+			'offset': 12-5, //10-5
+			'added': null,
+			'loaded': false,
+			'defaultLayout': 3,
+			'layout': 3,
 		};
 
 		this.fetchMore = this.fetchMore.bind(this);
@@ -147,10 +144,10 @@ export default class ContentGames extends Component {
 
 		var LoadMore = null;
 
-		if (error){
+		if ( error ) {
 			return <ContentError code="400">"Bad Request : Couldn't load games"</ContentError>;
 		}
-		else if(feed && feed.length > 0)
+		else if ( feed && feed.length > 0 )
 		{
 			var Games = feed.map(r => {
 				return <ContentItemBox node={r.node} user={props.user} path={props.path} noevent={props.noevent ? props.noevent : null} />;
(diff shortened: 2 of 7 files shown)


src/com/content-box/box.js
```diff
@@ -50,11 +50,7 @@ export default class ContentBox extends Component {
 
 	render( props, state ) {
 		props = Object.assign({}, props);
-
-		var node = props.node;
-		var user = props.user;
-		var path = props.path;
-		var extra = props.extra;
+		const {node, user, path, extra} = props;
```
povrazor: Hmm! I didn't realize this was working with the compiler. You might be on to something here. Collapsing many `var = props` lines in to one, I think I like it. :smile:

src/com/content-box/box.js
```diff
@@ -157,6 +153,8 @@ export default class ContentBox extends Component {
 					</div>
 				</ButtonLink>
 			);
+		} else if (props.placeHolder) {
```
povrazor: Prefer
```
}
else if (...) {
```

src/com/content-games/games.js
```diff
@@ -152,14 +156,28 @@ export default class ContentGames extends Component {
...
+				Games.push(<ContentItemBox placeHolder={true} />);
+			}
+
 			if ( !props.nomore /*|| added >= 10*/ ){
 				LoadMore = <ContentMore onclick={this.fetchMore} />;
 			}
 
+			const gridClass = `-columns-${layout}`;
+
 			return(
```
povrazor: Prefer
```
return (
```
(with a space between return and `(`).

I realize this is not a change from this set, just something to watch for.

src/com/content-games/grid-selector.js
```diff
@@ -0,0 +1,81 @@
...
+               <Dropdown
+                    items={options}
+                    onmodify={this.onSelectLayout}
+                    startExpanded={true}
+                />
+            );
+       }
+
+       return (
+           <div class={props.class ? props.class : 'grid-selector'}>
```
povrazor: Is this `class` correct? It might be, just want to double check that you want `grid-selector` if no class is specified. or do you always want grid selector (i.e. `class={cN(props.class, 'grid-selector')}`)?

And yes I'm trying to move us away from some of the older methods used for build class names, to the global `cN` function.

src/com/content-games/games.js
```diff
@@ -152,14 +156,28 @@ export default class ContentGames extends Component {
...
 
+			while (Games.length % layout !== 0) {
+				Games.push(<ContentItemBox placeHolder={true} />);
+			}
+
 			if ( !props.nomore /*|| added >= 10*/ ){
 				LoadMore = <ContentMore onclick={this.fetchMore} />;
 			}
 
+			const gridClass = `-columns-${layout}`;
```
povrazor: Template strings (``) aren't supported in IE11 (http://caniuse.com/#feat=template-literals), and I don't believe we have `dangerousTaggedTemplateString` enabled in our Buble config (because like the name suggests, there's a danger to it).

Prefer a simple string append. i.e. `"-columns-"+layout`  or `'-columns-'+layout`.

povrazor:
Not covered in the code review is that a bunch of the `/com/content-games` code should become moved up conceptually (i.e. it should apply to `item` content, not just `item/game` content (game is a specialization of an `item`)), and the idea of boxes renamed `cards`. 

These **don't** need to be addressed in this PR, but I just want to make note of them as a next step for this code after this.

src/com/content-games/grid-selector.js
local-minimum: You are right, that it is probably adding it to whatever else supplied is probably correct.

src/com/content-games/games.js
local-minimum: IE is why we can't have nice... but in this case non-formatted string was kind of fine too.

src/com/content-games/games.js
local-minimum: oops ofc

src/com/content-games/games.js
povrazor: Yeah this isn't so bad. `` is brand new. Now many people it's a feature (yet).

local-minimum:
As for making cards and generalizing, I'd prefer to use that as a different PR.
Except for the attending to the comments, I changed the icon svg to the cog since that seemed more to signify settings/config.

src/com/content-games/games.js
```diff
@@ -152,14 +156,28 @@ export default class ContentGames extends Component {
 				return <ContentItemBox node={r.node} user={props.user} path={props.path} noevent={props.noevent ? props.noevent : null} />;
 			});
 
+			while (Games.length % layout !== 0) {
```
povrazor: Could I get a comment here (line above) describing the cleverness going on (i.e. integer modulus is not a very common operation)?

src/com/content-games/grid-selector.js
```diff
@@ -0,0 +1,81 @@
...
+
+    componentDidMount() {
+       if (!this.state.selected) {
+          this.handleNewSelection(this.props.defaultLayout);
+       }
+    }
+
+    handleNewSelection(selected) {
+        const onChangeLayout = this.props.onChangeLayout;
+        this.setState({selected: selected});
```
povrazor: Please wrap member names in string quotes. i.e. `{'selected': selected}`. For consistency with JSON.

src/com/content-games/grid-selector.js
```diff
@@ -0,0 +1,81 @@
...
+import ButtonBase						from '../button-base/base';
+import SVGIcon 							from 'com/svg-icon/icon';
+
+export default class GridSelector extends Component {
+    constructor( props ) {
+		super(props);
+        this.state = {
+            expanded: false,
+            selected: null,
+        };
```
povrazor: Please wrap member names in string quotes. i.e. `{'expanded': false, ...}`. For consistency with JSON.

src/com/content-games/grid-selector.js
```diff
@@ -0,0 +1,81 @@
...
+        this.setState({selected: selected});
+        if (onChangeLayout) {
+            onChangeLayout(selected);
+        }
+    }
+
+    onToggleDropDown() {
+        console.log('toggle dropdown', this.state);
+        this.setState({expanded: !this.state.expanded});
+    }
```
povrazor: Please wrap member names in string quotes. i.e. `{'expanded': ...}`. For consistency with JSON.

src/com/content-games/games.js
```diff
@@ -152,14 +156,28 @@ export default class ContentGames extends Component {
...
+
+			return (
 				<div class={cN(Class, props.class)}>
 					{props.children}
-					<div class='content-boxes'>
+					<GridSelector
+						defaultLayout={defaultLayout}
+						onChangeLayout={
+							(gridLayout) => {
+								this.setState({layout: gridLayout,});
```
povrazor: Please wrap member names in string quotes. i.e. `{'layout': ...}`. For consistency with JSON.

povrazor:
Yes cards should definitely be a separate PR (I tried to say that, too many words I guess). Again just collecting thoughts since we're here.

src/com/content-games/games.js
local-minimum: Sure thing. Basically checks if the number of items evenly divides with number of columns

povrazor: Sounds good.

src/com/content-games/grid-selector.js
local-minimum: I'm so used to airbnb:s style where that is a crime

povrazor: Ha, no worries. I'm going to see if I can add that to the linter so it'll remember for us.

PR #1254, opened 2017-10-08
Simulator refactoring

sgstair:
Split up the simulator into a bunch of smaller files.
Also adding some more statistical stuff.


local-minimum:
breaking out the api seems nice, so it can mite easily be used for other purposes.

though tangential, we don't log any performance of the real api, do we? If we could, that would be really good stats for next ld

sgstair:
Most of the intent is to make the simulator code more approachable, I feel like it was a bit of a convoluted mess nobody else wanted to understand :) I'm not really sure if this will help, but it does make dev a little less painful (less scrolling involved)

Right, there's nothing on the API watching itself at the moment, I have been kicking around ideas about how to do this so we get the same kind of performance data in realtime from the site itself, to help with understanding what is expensive on the actual site. The simulated site is a pretty good proxy though.

local-minimum:
yeah but monitoring real site can help with better coping with load by scaling up where needed quickly. I don't know if it applies to ld, but at work we have prometheus to gather data and graphana to visualize and send off alarms

PR #1255, opened 2017-10-08
Fixes to notification code

sgstair:
May add some more later - I'm debugging the notifications page code which may have some issues with anonymous comments, but I need to finish some other things first.


local-minimum:
I wish the linter would complain about things not defined and things imported and not used... thanks for finding it

sgstair:
I suspect neither you nor pov saw this because it corresponds to a file that was deleted - so you'd only see this error on a clean rebuild.

local-minimum:
yeah, linter could have complained about that the import wasn't used

zwrawr:
Hit this issue as well, this fix is needed to build the source from scratch.

PR #1266, opened 2017-10-11
Games filtering

local-minimum:
image

This restructures the section on top of where games are shown so that all the filters are on dropdowns and there's a free text filter that will filter whatever games or content you have below based on words you write existing in either name or body of the items.

Note that the reason there's only one game below is thanks to the filter on word `case` and also, reason it gets ugly large is that this is based on master and not on the branch that fixes the layouts in the grid below.


povrazor:
Please fix the merge conflict. 😀

local-minimum:
Conflict should be resolved. I've never conflicted worth myself like that before. Also hit functionality issue with the new games grid drop down that I resolved. The dropdown component could probably do with some cleanup in the neat future thanks to my changes, but hey it works.

povrazor:
Okay. I'll take this as is. I don't know if it's 100% where I want it to be, but it's definitely an improvement over where things are right now. :smile:

local-minimum:
Yeah, I see it as first step and as soon as search is possible it should at least be rewired

PR #1267, opened 2017-10-11
PieCharts & Example Implimentaion

zwrawr:
Adds pie charts!!

image

### DIFF
--- a/src/com/content-stats/stats-event.js
+++ b/src/com/content-stats/stats-event.js
@@ -1,4 +1,4 @@
-import { h, Component } 				from 'preact/preact';
+import {h, Component} 				from 'preact/preact';
 
 import NavLink							from 'com/nav-link/link';
 
@@ -12,6 +12,9 @@ import ContentCommonBodyBy				from 'com/content-common/common-body-by';
 import ContentCommonBodyTitle			from 'com/content-common/common-body-title';
 import ContentCommonBodyAvatar			from 'com/content-common/common-body-avatar';
 import ContentCommonBodyMarkup			from 'com/content-common/common-body-markup';
+
+import PieChart							from 'com/visualization/piechart/piechart';
+
 
 import $Stats							from '../../shrub/js/stats/stats';
 
@@ -105,6 +108,26 @@ export default class ContentStatsEvent extends Component {
 				}
 
 				Data.push(<div class="-gap">Last Updated: {stats.timestamp}</div>);
+
+				let values = [
+					stats['grade-20-plus'],
+					stats['grade-15-20'],
+					stats['grade-10-15'],
+					stats['grade-5-10'],
+					stats['grade-0-5']-stats['grade-0-only'],
+					stats['grade-0-only']
+				];
+
+				let lables = [
+					'grade-20-plus',
+					'grade-15-20',
+					'grade-10-15',
+					'grade-5-10',
+					'grade-1-5',
+					'grade-0'
+				];
+
+				Data.push(<PieChart values={values} lables={lables}></PieChart>);
 
 				return (
 					<ContentCommon {...props} class={cN(Class)}>
--- a/src/com/content-stats/stats-event.js
+++ b/src/com/content-stats/stats-event.js
@@ -80,54 +80,64 @@ export default class ContentStatsEvent extends Component {
 
 			var Data = [];
 			if ( EventMode ) {
-				Data.push(<div><span class="-title">Signups:</span> <span class="-value">{stats.signups}</span></div>);
-				Data.push(<div><span class="-title">Unique Authors:</span> <span class="-value">{stats.authors}</span></div>);
+				Data.push(<div class="-gap"><span class="-title">Number of Users:</span></div>);
+				Data.push(<div><span>Signups:</span> <span class="-value">{stats.signups}</span></div>);
+				Data.push(<div><span>Unique Authors:</span> <span class="-value">{stats.authors}</span></div>);
 
 				if ( EventMode >= 5 ) {
-					Data.push(<div class="-gap"><span class="-title">Submissions:</span> <span class="-value">{stats.jam+stats.compo+stats.craft}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Jam:</span> <span class="-value">{stats.jam}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Compo:</span> <span class="-value">{stats.compo}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Craft:</span> <span class="-value">{stats.craft}</span></div>);
-					Data.push(<div><span class="-title">Unfinished:</span> <span class="-value">{stats.unfinished}</span></div>);
-					Data.push(<div><span class="-title">Unpublished:</span> <span class="-value">{stats.unpublished}</span></div>);
-					Data.push(<div><span class="-title">Warmups:</span> <span class="-value">{stats.warmup}</span></div>);
+					Data.push(<div class="-gap"><span class="-title">Number of Submissions:</span> <span class="-value -title">{stats.jam+stats.compo+stats.craft}</span></div>);
+					let entries = [
+						stats.compo,
+						stats.jam
+					];
+					let entrienames = [
+						'Jam',
+						'Compo'
+					];
+					Data.push(<PieChart values={entries} lables={entrienames}></PieChart>);
+
+					Data.push(<div><span>Unfinished:</span> <span class="-value">{stats.unfinished}</span></div>);
+					Data.push(<div><span>Unpublished:</span> <span class="-value">{stats.unpublished}</span></div>);
+					Data.push(<div><span>Warmups:</span> <span class="-value">{stats.warmup}</span></div>);
 				}
 				if ( EventMode >= 6 ) {
-					Data.push(<div class="-gap"><span class="-title">Ratings:</span></div>);
-					Data.push(<div class="-indent"><span class="-title">20+:</span> <span class="-value">{stats['grade-20-plus']}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">15-20:</span> <span class="-value">{stats['grade-15-20']}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">10-15:</span> <span class="-value">{stats['grade-10-15']}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">5-10:</span> <span class="-value">{stats['grade-5-10']}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">0-5:</span> <span class="-value">{stats['grade-0-5']}</span></div>);
-					Data.push(<div class="-gap"><span class="-title">With 0 ratings:</span> <span class="-value">{stats['grade-0-only']}</span></div>);
-					Data.push(<div class="-gap"><span class="-title">Totals:</span> <span class="-value">{stats.game}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Games:</span> <span class="-value">{stats.game}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Crafts:</span> <span class="-value">{stats.craft}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Demos:</span> <span class="-value">{stats.demo}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Tools:</span> <span class="-value">{stats.tool}</span></div>);
+					Data.push(<div class="-gap"><span class="-title">Number of Ratings Recived (Per Game):</span></div>);
+					let grades = [
+						stats['grade-20-plus'],
+						stats['grade-15-20'],
+						stats['grade-10-15'],
+						stats['grade-5-10'],
+						stats['grade-0-5']-stats['grade-0-only'],
+						stats['grade-0-only']
+					];
+					let gradenames = [
+						'20-plus',
+						'15-20',
+						'10-15',
+						'5-10',
+						'1-5',
+						'0'
+					];
+					Data.push(<PieChart values={grades} lables={gradenames}></PieChart>);
+
+					Data.push(<div class="-gap"><span class="-title">Types of Entries:</span></div>);
+					let types = [
+						stats.game,
+						stats.craft,
+						stats.demo,
+						stats.tool
+					];
+					let typenames = [
+						'Games',
+						'Crafts',
+						'Demos',
+						'Tools'
+					];
+					Data.push(<PieChart values={types} lables={typenames}></PieChart>);
+
 				}
 
 				Data.push(<div class="-gap">Last Updated: {stats.timestamp}</div>);
-
-				let values = [
-					stats['grade-20-plus'],
-					stats['grade-15-20'],
-					stats['grade-10-15'],
-					stats['grade-5-10'],
-					stats['grade-0-5']-stats['grade-0-only'],
-					stats['grade-0-only']
-				];
-
-				let lables = [
-					'grade-20-plus',
-					'grade-15-20',
-					'grade-10-15',
-					'grade-5-10',
-					'grade-1-5',
-					'grade-0'
-				];
-
-				Data.push(<PieChart values={values} lables={lables}></PieChart>);
 
 				return (
 					<ContentCommon {...props} class={cN(Class)}>
--- a/src/com/visualization/piechart/piechart.js
+++ b/src/com/visualization/piechart/piechart.js
@@ -21,11 +21,6 @@ export default class PieChart extends Component {
         return n_values;
     }
 
-    // https://stackoverflow.com/a/1152508
-    randomColor() {
-        return '#' + (0x1000000 + (Math.random() * 0xFFFFFF)).toString(16).substr(1, 6);
-    }
-
     render( props ) {
         console.log(props);
 
@@ -41,14 +36,14 @@ export default class PieChart extends Component {
 
         for (var i = 0; i < n_values.length; i++) {
 
-            let color = this.randomColor();
-            let style = "background-color:"+color+";";
+            let color = 1 + ( i % 6 );
+            let ledgendclass = "shape-circle ledgend_color_"+color;
 
             Segments.push(<PieSegment angle={n_values[i]} offset={offset} color={color}></PieSegment>);
             Ledgend.push(
                 <li>
-                    <span class="shape-circle" style={style}></span>
-                    {lables[i]} ({Math.round(n_values[i] * 100) / 100}%)
+                    <span class={ledgendclass}></span>
+                    <p>{lables[i]} ({values[i]} / {Math.round(n_values[i] * 100) / 100}%)</p>
                 </li>
             );
 
--- a/src/com/visualization/piechart/piechart.js
+++ b/src/com/visualization/piechart/piechart.js
@@ -1,0 +1,71 @@
+import {h, Component} 				from 'preact/preact';
+
+import PieSegment					from 'com/visualization/piechart/piesegment';
+
+export default class PieChart extends Component {
+
+	constructor( props ) {
+		super(props);
+
+    }
+
+    normalize ( values ) {
+        let total = 0;
+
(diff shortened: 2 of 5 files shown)


povrazor:
Oh that's cool looking. :smile: I'll try to get to this later.

src/com/content-stats/stats-event.js
```diff
@@ -77,31 +80,61 @@ export default class ContentStatsEvent extends Component {
 
 			var Data = [];
 			if ( EventMode ) {
-				Data.push(<div><span class="-title">Signups:</span> <span class="-value">{stats.signups}</span></div>);
-				Data.push(<div><span class="-title">Unique Authors:</span> <span class="-value">{stats.authors}</span></div>);
+				Data.push(<div class="-gap"><span class="-title">Number of Users:</span></div>);
```
local-minimum: A rather smallish thing, but I would prefer these three lines to be a single JSX (i.e. have outer div) and be assigned to a variable with a helping name.

`ShowIntroStats = <Common>your lines </Common>;`

I'm not sure exactly what of the already defined components are most correct but should be easy to spy on other files. 

Then at the top you would have some `let ShowIntroStats = null`

This is at least the general pattern I've seen in the other ui files.

Using arrays feels like it should be reserved for preparing a list of near identical things that could be done in a `forEach` or similar.

src/com/content-stats/stats-event.js
```diff
@@ -77,31 +80,61 @@ export default class ContentStatsEvent extends Component {
...
-					Data.push(<div class="-indent"><span class="-title">Jam:</span> <span class="-value">{stats.jam}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Compo:</span> <span class="-value">{stats.compo}</span></div>);
-					Data.push(<div class="-indent"><span class="-title">Craft:</span> <span class="-value">{stats.craft}</span></div>);
-					Data.push(<div><span class="-title">Unfinished:</span> <span class="-value">{stats.unfinished}</span></div>);
-					Data.push(<div><span class="-title">Unpublished:</span> <span class="-value">{stats.unpublished}</span></div>);
-					Data.push(<div><span class="-title">Warmups:</span> <span class="-value">{stats.warmup}</span></div>);
+					Data.push(<div class="-gap"><span class="-title">Number of Submissions:</span> <span class="-value -title">{stats.jam+stats.compo+stats.craft}</span></div>);
+					let entries = [
+						stats.compo,
+						stats.jam
```
local-minimum: I'm a fan of trailing comma on last item but not sure what Mike feels about it.

Reason is that additions to arrays don't need to change as  many lines (i.e. add comma to previous last line)

povrazor: I do both. Typically trailing commas show up in places that there's an expectation (or actuality) that the contents will change.

src/com/content-stats/stats-event.js
povrazor: Heh, the pushes initially were actually me. :laughing:

This file was written in a hurry, and each line is basically a copy+paste with a different data field as I exposed them with the API. I was more concerned with roughing it in than anything. There was no rhyme or reason. I just kept adding new stats as I had them.

povrazor: On a side node, though I don't mind the style of the `let ShowBlah = <div />;`, I can't say I'm 100% happy with the word `Show`. I've been thinking the same thing any time I've seen a member variable named `Class` (first character in caps), thinking it should be `Classes` or something else. Anyway, not really related to this issue.

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+        let percentages = values.map( ( v ) => ( (100 * v) / total ) );
+
+        return percentages;
+    }
+
+    render( props ) {
+
+        if ( !(props && props['lables'] && props['values'])) {
+            console.warn('PieChart was created with invalid props', props);
+            return <div> No Data! </div>;
```
povrazor: Prefer
```
return <div>No Data!</div>;
```

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+
+        // scale values so there sum is 100%
+        let percentages = values.map( ( v ) => ( (100 * v) / total ) );
+
+        return percentages;
+    }
+
+    render( props ) {
+
+        if ( !(props && props['lables'] && props['values'])) {
```
povrazor: Super minor (probably a typo), but prefer
```
if ( !(props && props['lables'] && props['values']) ) {
```
noticing the added space at the end separating the `))`.

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+        let lables = props.lables;
+        let values = props.values;
+
+        let percentages = this.convertToPercentage(values);
+
+        let Segments = [];
+        let Ledgend = [];
+
+        let offset = 0;
+        for (var i = 0; i < percentages.length; i++) {
```
povrazor: Prefer
```
for ( var i = 0; i < percentages.length; i++ ) {
```

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
+import {h, Component} 				from 'preact/preact';
+
+import PieSegment					from 'com/visualization/piechart/piesegment';
+
+export default class PieChart extends Component {
+
+	constructor( props ) {
```
local-minimum: constructor that just does super should not be needed

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+        let Segments = [];
+        let Ledgend = [];
+
+        let offset = 0;
+        for (var i = 0; i < percentages.length; i++) {
+
+            let color = 1 + ( i % 6 );
+            let ledgendclass = "shape-circle ledgend_color_"+color;
+
+            Segments.push(<PieSegment angle={percentages[i]} offset={offset} color={color}></PieSegment>);
```
povrazor: If there's no plans to put anything inside a set of `<PieSegment></PieSegment>` tags, one could just use `<PieSegment />`. If there is however, just ignore me. :laughing:

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+
+export default class PieChart extends Component {
+
+	constructor( props ) {
+		super(props);
+
+    }
+
+    // Converts an array of number to an array of percentages
+    convertToPercentage ( values ) {
```
local-minimum: could be static or own function outside of class for easier reuse

src/com/visualization/piechart/piechart.less
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    & .ledgend {
+        flex: 1;
+        padding-left: 15px;
+        padding-right: 15px;
+        align-self: center;
+          
+        li {
```
povrazor: Prefer
```
& li
```

src/com/visualization/piechart/piechart.less
```diff
@@ -0,0 +1,69 @@
...
+            
+            & .shape-circle {
+                display: inline-block;
+                vertical-align: middle;
+                width: 32px;
+                height: 32px;
+                border-radius: 50%;
+            }
+
+            p {
```
povrazor: Prefer
```
& p
```

src/com/visualization/piechart/piechart.less
```diff
@@ -0,0 +1,69 @@
...
+.piechart_color_4 {stroke: @COL_AB;}
+.piechart_color_5 {stroke: @COL_BC;}
+.piechart_color_6 {stroke: @COL_CA;}
+
+.ledgend_color_1 {background-color: @COL_A;}
+.ledgend_color_2 {background-color: @COL_B;}
+.ledgend_color_3 {background-color: @COL_C;}
+.ledgend_color_4 {background-color: @COL_AB;}
+.ledgend_color_5 {background-color: @COL_BC;}
+.ledgend_color_6 {background-color: @COL_CA;}
```
povrazor: File should end with newline (linter should have caught this)

src/com/visualization/piechart/piesegment.js
```diff
@@ -0,0 +1,34 @@
...
+
+	constructor( props ) {
+		super(props);
+
+    }
+
+    render( props ) {
+
+        // it's valid for offset to be zero so we have to check it against undefined.
+        if ( !(props && props['angle'] && props['offset'] != undefined && props['color'])) {
```
povrazor: Prefer a space between the `))`'s at the end (to match the `if ( ` at the front).

src/com/visualization/piechart/piesegment.js
```diff
@@ -0,0 +1,34 @@
...
+            console.warn('PieSegment was created with invalid props', props);
+            return;
+        }
+
+        let angle = props.angle;
+
+        // drawing a segment of 0 width causes artifacting so bail out.
+        if (angle == 0) {
```
povrazor: Prefer
```
if ( angle == 0 ) {
```

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+        // scale values so there sum is 100%
+        let percentages = values.map( ( v ) => ( (100 * v) / total ) );
+
+        return percentages;
+    }
+
+    render( props ) {
+
+        if ( !(props && props['lables'] && props['values'])) {
+            console.warn('PieChart was created with invalid props', props);
```
local-minimum: consider

```
const {labels, values} = props;
if (! (labels && values) )
```

Or even replace props in the 

```
render ({labels, values})
```

if these are the only two fields from props used

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+        let percentages = this.convertToPercentage(values);
+
+        let Segments = [];
+        let Ledgend = [];
+
+        let offset = 0;
+        for (var i = 0; i < percentages.length; i++) {
+
+            let color = 1 + ( i % 6 );
+            let ledgendclass = "shape-circle ledgend_color_"+color;
```
local-minimum: I think it would be nicer to use the global class name formatting function `cN`

```
cN("shape-circle", "legend-color-"+color, props.legendClass);
```
since then ready to be extended if we want per components optionally passing in extra definitions

src/com/visualization/piechart/piechart.js
local-minimum: I'd prefer
```
percentages.forEach( (percentage, i) => { ...}
```

src/com/visualization/piechart/piechart.js
```diff
@@ -0,0 +1,69 @@
...
+
+            Segments.push(<PieSegment angle={percentages[i]} offset={offset} color={color}></PieSegment>);
+            Ledgend.push(
+                <li>
+                    <span class={ledgendclass}></span>
+                    <p>{lables[i]} ({values[i]} / {Math.round(percentages[i] * 100) / 100}%)</p>
```
local-minimum: There should be a function to round to two decimal places in Math or on the number object

src/com/visualization/piechart/piesegment.js
```diff
@@ -0,0 +1,34 @@
+import {h, Component} 				from 'preact/preact';
+
+export default class PieSegment extends Component {
+
+	constructor( props ) {
```
local-minimum: Not needed (i think )

src/com/visualization/piechart/piesegment.js
local-minimum: Also consider my comment for the other file

povrazor:
Okay. You're a bit ahead of the rest of us by putting this in `com/visualization/piechart/` (we haven't started moving things in to subfolders yet, but it will happen). I more want to have a mini discussion.

Main thing: I'm not sure I like the name `visualization`. 

There's disorder in `com`. Some of it follows a reasonable structure (`views` are part of the page, `content`/`header`/`sidebar` is data that goes in those respected views, `input` are things that receive input *). 

On my TODO list is to eventually rename `ButtonBase`, `ButtonLink`, `NavLink` and a few others in to types that are prefixed with `input`, because that is what they do. We have other types that are purely visual, like `SVGIcon`, `Img2`, `nav-spinner`, and `label-yesno`. I think piechart belongs alongside them. It's a component used to make up a page.

The problem is I don't know what to name these purely visual/display only elements.

Suggestions welcome. :wink:

src/com/visualization/piechart/piechart.js
povrazor: Actually I'd prefer that we avoid destructuring inside function definitions. By destructuring you lose the props variable, which is useful for a variety of things.

src/com/visualization/piechart/piechart.js
povrazor: classic for loops are a construct found in both JavaScript and PHP. 

Also: https://jsperf.com/for-vs-foreach/37

src/com/visualization/piechart/piechart.js
povrazor: It's a common stub. If something does ever need to be done during construction, it's already there waiting.

src/com/visualization/piechart/piechart.js
povrazor: When it's something this small, I prefer to wait until I actually need something a 2nd time before relocating. If we had a 2nd chart type, different story. Right now it's just the one.

src/com/visualization/piechart/piechart.js
povrazor: Yeah, cN is new and slowly replacing all class merging fancyness. If you Google the classNames library/function, it's that, but I abbreviated it to cN.

src/com/visualization/piechart/piechart.js
povrazor: Yeah MyNumber.toFixed(2) might be more elegant.

src/com/visualization/piechart/piechart.less
povrazor: Oh, and I think you spelled legend wrong. 😉

src/com/visualization/piechart/piechart.js
zwrawr: So the above implementation and number.toFixed()  are slightly different.

consider 10%
The current implementation outputs 10% where as toFixed(2) outputs 10.00%

src/com/visualization/piechart/piechart.js
zwrawr: Changed to cN implementation.

src/com/visualization/piechart/piechart.js
zwrawr: Keeping it as `render(props)`

src/com/visualization/piechart/piechart.js
zwrawr: If i add more chart types i'll break this (and probably all the legend stuff) out into different files that can be used by all of the charts/graphs ect.

src/com/visualization/piechart/piesegment.js
zwrawr: Added spacing

src/com/visualization/piechart/piesegment.js
zwrawr: Added spacing

src/com/visualization/piechart/piechart.less
zwrawr: Oops, dyslexia strikes again. :)

src/com/visualization/piechart/piechart.js
zwrawr: Changed to a self closing tag

src/com/visualization/piechart/piechart.less
zwrawr: Added &

src/com/visualization/piechart/piechart.less
zwrawr: Added &

src/com/visualization/piechart/piechart.js
zwrawr: Keeping it as the classic for, it's the most common loop type when you need the index.

src/com/visualization/piechart/piechart.js
zwrawr: Removed extra spacing

src/com/visualization/piechart/piechart.js
zwrawr: Added spacing

src/com/content-stats/stats-event.js
zwrawr: Not adding trailing spaces, we probably wont run more sub-events any time soon.
And eventually this may need to be generalised for non LD events so it will probably pull the list from the api at that point.

src/com/content-stats/stats-event.js
zwrawr: I think the array is the cleanest way to do it in this case, otherwise we would have a dozen separate variables.

zwrawr:
> I would also consider allowing styling radii of the circle segment 
Why would we need this? Wouldn't that just make the site look inconsistent?

> Main thing: I'm not sure I like the name visualization.
Yea, I didn't really a good idea for what to call them. But it makes sense to group any chart, graphs ect. together somewhere.

src/com/content-stats/stats-event.js
local-minimum: Again just me and airbnb style guide, and I think it is a rule that makes sense. But not a strong opinion.

src/com/visualization/piechart/piechart.js
local-minimum: to me this is a maths helper function that has very little to do with the graphs and could be in separate file that doesn't deal with components and you import when needed. 

bad name would be `helpers.js`  but it explains the purpose. (To me that's where the `cN` function belongs too rather than the global state.)

src/com/content-stats/stats-event.js
local-minimum: I don't agree, as far as I see it about 5 or less are needed.

```
ShowSingups
ShowCategory
ShowVoting
```

Each would correspond to several of current `push` since they are logical blocks anyhow and could just as well have a div around themselves.

src/com/visualization/piechart/piechart.js
local-minimum: but dead code... ;(

local-minimum:
The radii settings, might not be so important. Especially now. But if the chart found usage in different context likr a popup, the size constraints on the area available could encourage different settings.

As for names, I'm not sure, but there are some slight similarities with `box`/`card` in that it is mostly graphics with minor text. I would like to trust `cards` to be ready to be flexed though. I think.

I think `graph` is such a strong concept that I'd like to see it included from start whatever we settle on. And on a side note. That initial data with number of signups and number of authors could be in a bar graph.

Other cool stats would be number of posts and comments madewith the event in parent-tree. Preferably split per node type.

Not at all suggesting it should be part of this PR

sgstair:
As for a category name for a class of display / output components, I can think of a few suitable names - 'render', 'widget' (though has input and output connotations), 'display', 'output'. Nothing particularly stands out as brilliant though.

src/com/visualization/piechart/piechart.js
povrazor: If we're 100% sure it wont be used, then absolutely.

povrazor:
> Nothing particularly stands out as brilliant though.

Yeah this is my dilemma. Going the `widget` route is an option too. I don't mind up-rooting my `input` plans if it meant all "things" used to make a page were in one place.

src/com/content-stats/stats-event.js
povrazor: Yeah it's not a bad practice (I could probably be sold on it). What's especially good about it is you don't end up accidentally changing a line just to add a comma when things are inserted, so the patches I see are cleaner. I just wish it was legal JSON to do this (it's not, most parsers will error if you do this). TBD. If you keep poking me about this I might convert. :laughing: 

But back to the point: This stats page is very "just there" at the moment. It's not well laid out or structured. My approach was more "lets get all the stats we care about on the page, then decide the best layout".

PR #1269, opened 2017-10-12
Shrub tags

local-minimum:
Idea is to allow UI to query tags based on their names.
Matching is done in case independent and any initial `#` is removed from input.
All tags containing pattern are returned.
Uses existing cache.

Example REPL use:
```
IN:> $Tag.GetContains('y').then(r=>console.log(r.tag.map(t=>t.name)));
OUT:> Array [ "kitty", "Gameboy" ]
```


povrazor:
Ah, for autocompletion sort of thing?

local-minimum:
Yes do a dropdown/up? while typing in the search/filter bar.

Issue #345, opened 2016-12-02
Theme Slaughter: Show stats!


karai17:
I would like to know how many themes there are, how many I've voted on, how many I've Yes, No and Flagged, and a general breakdown of where I sit within the rest of the community. :)

povrazor:
Good idea.

karai17:
Yeah, I remember those and I really liked them. Simple text stats would be good enough for now though.

local-minimum:
So what does grade mean here if we are talking theme-stats?

povrazor:
What are you using for the chart?

zwrawr:
The example above is on the event stats page. It's the number of games that got x number of rating.

The chart is native, seemed over kill to add a charting/graphing library for a few piecharts. (And i wanted something complex to work on so i could relearn some css)

If we end up needed lots of graphs then we'll probably need a library, but right now we only need piecharts.

povrazor:
Cool. :smile:

local-minimum:
That explains why I was a bit confused, the data being for the other stats page. Still very nice.

zwrawr:
Yea, i miss remembered what the issue was.

zwrawr:
Closed By #1275

PR #1275, opened 2017-10-13
Adds piechart to theme slaughter page

zwrawr:
As suggested in #345

### DIFF
--- a/src/com/content-event/event-slaughter.js
+++ b/src/com/content-event/event-slaughter.js
@@ -188,38 +188,46 @@ export default class ContentEventSlaughter extends Component {
 		return Object.keys(this.state.ideas).map(this._renderMyIdea);
 	}
 
-	renderBody( state /*{current, votes, votes-left,  ideas, done, error}*/ ) {
+	renderBody( state ) {
 
-		console.log(state.votes);
-		let kept = Object.values(state.votes).reduce((t, v) => {
-			return t+v;
-		});
-
-		var values = [
-			kept,
-			Object.keys(state.votes).length - kept,
-			state['votes-left']
-		];
+		let seen = Object.keys(state.votes).length;
 
 		var labels = [
-			'kept',
 			'slaughtered',
+			'kept',
 			'left'
 		];
 
+		var values = [
+			0,
+			0,
+			state['votes-left']
+		];
+
+		if ( seen != 0 ) {
+			Object.values(state.votes).forEach((v) => {
+				if ( v == 0 ) {
+					values[0]++;
+				}
+				else if ( v == 1 ) {
+					values[1]++;
+				}
+				// else if v=2 it's flagged
+			});
+		}
+
 		var StatsAndDetails = (
-			<div>
+			<div class="history">
 				<h3>Recent Themes</h3>
 				{this.renderRecentQueue()}
-				<PieChart values={values} labels={labels} />
 			</div>
 		);
 
 		if ( state.done ) {
 			return (
-				<div>
+				<div class="-stats">
 					<div>Wow! {"You're totally done!"} Amazing! You slaughtered {Object.keys(state.votes).length} themes!</div>
-					{StatsAndDetails}
+					<PieChart values={values} labels={labels} />
 				</div>
 			);
 		}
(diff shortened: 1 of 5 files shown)


src/com/content-event/event-slaughter.js
```diff
@@ -178,56 +180,85 @@ export default class ContentEventSlaughter extends Component {
...
 		);
 
-		if ( done ) {
+		if ( state.done ) {
 			return (
-				<div>
-					<div>Wow! {"You're totally done!"} Amazing! You slaughtered {Object.keys(votes).length} themes!</div>
-					{StatsAndDetails}
+				<div class="-stats">
+					<div>Wow! {"You're totally done!"} Amazing! You slaughtered {Object.keys(state.votes).length} themes!</div>
```
local-minimum: This is probably not yours, but why is that a string in curlies?

src/com/visualization/piechart/piechart.js
```diff
@@ -23,12 +23,12 @@ export default class PieChart extends Component {
 
     render( props ) {
 
-        if ( !(props && props['lables'] && props['values']) ) {
+        if ( !(props && props['labels'] && props['values']) ) {
```
local-minimum: why not `props.labels?

src/com/content-event/event-slaughter.js
zwrawr: I think it might be so that the punctuation doesn't need to be escaped, but that is only a guess?

src/com/visualization/piechart/piechart.js
zwrawr: I cannot be certain that someone has passed in labels={} to the component.

src/com/content-event/event-slaughter.js
local-minimum: oh I think it is "worse" than that, come to think about it. Probably only to make editor not falsly misinterpret it though it is perfectly valid

src/com/visualization/piechart/piechart.js
local-minimum: `props.label` should still evaluate as false in that case.

src/com/content-event/event-slaughter.js
povrazor: Yeah I'm using an editor that really doesn't understand JSX. Now that I'm sharing though, we should disregard stuff that's broken in my editor and go with what looks right.

src/com/visualization/piechart/piechart.js
povrazor: Yeah this is one of those JS silly things. As long as we're sure the parent (props) is an object, you can safely use `props.labels` and it wont die a horrible death.

src/com/visualization/piechart/piechart.less
```diff
@@ -21,6 +21,7 @@
```
povrazor: This might seem like a weird suggestion, but perhaps `.piechart-figure` should be named `.chart-pie` (as there are potentially other types of charts). I'm assuming this is the container that contains all the stuff.

Since everything else is a child (never used outside the pie chart code), the classes should be named with a `-` prefix. i.e. `& .-piechart`, but there's more.

A `.chart-pie` likely only contains a single chart, so `& .-chart` is an acceptable class name for the child.

Finally, it's likely `& .-chart` is always a direct descendant of `.chart-pie`, so it should include a `>`.

i.e. `& > .-chart {`

src/com/visualization/piechart/piechart.less
```diff
@@ -21,6 +21,7 @@
 
         & .piechart-svg {
```
povrazor: This is then a direct decedent of the `& > .-chart`, and best I can tell it will be the only SVG child of it, so you can omit the class and go with `& > svg {`

src/com/visualization/piechart/piechart.less
```diff
@@ -21,6 +21,7 @@
...
             & .piechart-segment {
```
povrazor: Finally, since the piechart relationship is a given, we can git explicit again. `& .-segment {`

I omitted the `>` here so to not limit how you structure the things inside the SVG. If you are certain it's a direct descendent, you can include it.

src/com/content-event/event-slaughter.js
zwrawr: Removed enclosing `{""}`

src/com/visualization/piechart/piechart.js
zwrawr: JS is weird, changed.

Issue #1194, opened 2017-08-24
Provide a mechanism to get a short list of random games


sgstair:
Ideally, weighted with the smart rating (where smart greater than zero)

So people can use this for on-the-spot random game selection for rating (I'd use the heck out of a feature like this, rather than purely browsing the smart ratings)

Efficient implementation may require caching smart rating + node id in an additional side table, in the magic task.

povrazor:
Related to #954

local-minimum:
I might look into making a widget for displaying games according to results from api call, basically a simplified feed view of the games view without options on how to filter results.

povrazor:
The existing code that displays games can be improved for this purpose (that's how the "pinned" news works). 

On this subject, the box view needs improvement. Right now it's limited to 3 items wide inside the content view, and when a non-3 number of items is returned you get a jumbo item (due to there being no filler items). This needs to support way more choices (1 wide, 2 wide, 4 wide, 5 wide, etc).

local-minimum:
Yeah I figured, doing the suggested widget would trigger reworking the existing. So many fun UI things to work on.

Issue #1193, opened 2017-08-24
Consider: adapting smart filter to boost value of highly rated games


sgstair:
The argument is, we should get more visibility on games that are performing very well before voting closes, so we can get better statistical certainty on  the true value for their ratings.

One concern is, you could potentially use this change to infer something about your game's ratings before the voting period is over. One possible implementation to work around this issue would be to issue a fixed smart rating boost for games with any rating category value above some fixed level.

sgstair:
(This was suggested by cerno-b in the thread at https://ldjam.com/events/ludum-dare/39/false-game/results-for-ludum-dare-39)

povrazor:
I had this same thought some time ago, but I didn't have a great concept for it.

The main issue is that scores are one of the magic data's. Calculating them isn't a cheap operation (yet). That's why other magic stuff only does 500 at a time, so we're not bogging things down. Also anything we include in magic is potentially public.

If a factor is included in the smart filter, given every other piece of public data, it's likely folks will be able to detect which games are being bumped by this. 

I kinda think this is something that should maybe be revisited later, once we've done some things that improve how active voters are. I don't have stat's but I'm certain that voting rates reeeeealy slow down after the first week. So switch some mode on like this might not be very effective, unless more people are actively voting during this time.

Cerno-b:
I thought about this a bit more. One solution would be to compute the smart filter order every minute and store the results in the database. Then you would only have to sort by order.

A way to prevent people from guessing they are boosted because of high ratings would be to randomize the smart filter. Instead of treating the smart score as a fixed value to be sorted it would be used as a weight for randomization. You throw the dice every 5 minutes and games with 2 times the score of other games will be twice as likely to receive a high smart score. Not sure whether it's okay to keep the list static for so long.

A less intrusive way would be to leave everything as it is and pick a game randomly that is eligible for a boost. That game would then get a bonus on smart score that decreases over time. It will show up on the smart filter page 1 or 2 for about 10 minutes and then slowly drop down over time. There would be boost candidates sliding through the smart page for a short time which would limit people from realising a specific game being pushed.

It is probably a bit fiddly to set up but finding a covert way to give high rating/low vote games a boost in votes will get rid of a lot of animosity when the results are revealed

Cerno-b:
Maybe the boost drop off is a bit too much effort though. A simple "show a random eligible game on the smart filter list page 2 for 15 minutes" may be easier to set up.

sgstair:
[...]

PR #1277, opened 2017-10-14
Random game picker API

sgstair:
This is the API side for #1194

This PR introduces a new API to get a list of random games. It's been designed to not hit the database a lot (mainly caches the data in APCu) and should be acceptably fast. Details below:
* New random games GET API (~~`/vx/randomgames/get`~~ `/vx/random/game/get`) can query for up to 50 random games, returned as a node id list in `games`. There's also another returned value `listenabled` which will be true and the games list non-empty only when the event is in its Voting or Complete phases.
* Query with `?count=n` to get a specific number of games, otherwise the default is 5
* Provides a mechanism to filter, you can query for ~~`/vx/randomgames/get/compo`~~ `/vx/random/game/get/compo` or `/jam` (other values are meaningless currently)
* Data from the last simulation run shows calls take on average `1.02ms` (considerably better than most calls in the API), and max of `3.9ms` (when generating the cached structure, given the 40 or so games on the simulation event)

As for operation:
* `node_randomgames.php` will, on request, generate a structure containing a list of all games in the current event, in a minimal format. It does two custom DB queries to retrieve only as much data as it needs. It then computes a list of games and weights, games with a high enough weight go into a weighted random list, any other games go into an unweighted list which games are picked from randomly if the weighted random list runs out.
* Once the structure is generated, it's cached for 3 minutes to reduce DB load. This structure is invalidated at the end of magic.php so fresh data will immediately be computed from the updated smart ratings.
* From the generated data structure, the code uses a weighted random selection algorithm to pick games based on their weight. In the future we can boost these weights based on private information to help ensure the games with highest scores will get more attention, without revealing this information (provided that people use this mechanism to select games) - This aspect is #1193. If the system runs out of games with weights, it will pad the list with unweighted games.

Unrelated changes:
* Simulator is now adding more markdown styles into its text.


public-api/vx/randomgames.php
```diff
@@ -0,0 +1,51 @@
...
+		
+		// If there is a user id, find the user's games and exclude them
+		$excludegames = [];
+		if ( $user_id = userAuth_GetID() ) {		
+			$userauthoredlinks = nodeLink_GetByKeyNode("author", $user_id);
+			foreach( $userauthoredlinks as $authorlink ) {
+				// This will only go one way, but guard anyway.
+				if ( $authorlink['b'] == $user_id ) { $excludegames[] = $authorlink['a']; }
+			}
+		}
```
local-minimum: This does not exclude games voted on right? That would be a nice filter to have too to increase the usefulness.

sgstair: It excludes games you created :) That's a good point though, and I can certainly add that.

sgstair:
Fixed both issues that were brought up.
* `/randomgames/get` no longer includes games that you have voted on
* New API `/grade/getgames` (or `/grade/getgames/[event node ID]`) that returns `event_id` and `games` as a list of node IDs your user has voted on.

local-minimum:
I'm not sure what the comment list is in this context, but my plan was to soonish add a games I have voted on on the featured event. But I figured it should be close to the notifications. So /home/votehistory or something

sgstair:
re: comment list - I mean I have long planned to build a page to list comments you've posted. Not exactly the same as games you've commented on,

Noojuno:
It might be better to make it `/vx/random/get/` with `type`, `subtype`, and `subsubtype` arguments in case it's wanted for any other type of node that isn't a game. I think the more generic an api call for this sort of thing the better.

local-minimum:
(comments list I'd like too. `/home/history/comments` I'd rather place it cuz it isn't about seeing what someone else had commented cross the site, but to see you own.)

sgstair:
@noojuno It's very very not generic; I can't imagine returning random other kinds of nodes, that's why I chose to name it like this.

@local-minimum I think it's useful for the comment history (minus anonymous comments) to be visible to other users also.

sgstair:
still thinking over the rename of /randomgames, but I'm leaning towards the opinion that it's appropriate. Of course if @povrazor has an opinion I may have to reconsider :P

Noojuno:
@sgstair I could see potential scenarios where a random user or a random video (if they're implemented in future) could be useful. Same as a random tag or similar

povrazor:
Definitely de-game it.

sgstair:
Renamed the endpoint to /random/game - If you mean something further please clarify. As it is the code is necessarily very game specific. If we want to do a more generic random picker in the future, I think we have some more design to do.

povrazor:
Yeah, this is fine for now.

PR #1280, opened 2017-10-20
Preparations for Tag autocomplete options popup

local-minimum:
Several uses of the dropdown actually only needed/wanted the option-list aspect, so I made that a separate component here.

Tested to work with:
* `/games` (the grid number of columns thing, options-list),
* user's game editing links (tags, actual dropdown), 
* notifications (options-list)
* games-filters (actual dropdown)

Fixes so that notifications dropdown will hide if you click on going to the notifications page.

## Edit

Now also fixes UX issue with dropdowns not collapsing if reselecting same option, also some cleaning of code at the same time.


povrazor:
I still don't understand why a `navChange` is needed.

local-minimum:
It is usable but much less flexible. But I realized how to fix the issue though code will bocome more complex. For whatever option is selected I will need to swapout the ButtonLink for my own div, since I need to be able to else on the fly toggle when the button should navchange and when not.

Button active in both cases.

local-minimum:
anyways, I fixed the issue now. Everything should be fine.

povrazor:
Just for reference, though it may not seem that way (style bugs), ButtonBase and ButtonLink are supposed to be interchangeable (one for a clickable thing that does an action, other for a clickable link).

local-minimum:
Yeah, I managed to work with it.

Let me know if you have any issues with the code

povrazor:
Ah, and I was just about to look at this 'cause I literally need
autocompletion next.

local-minimum:
I think this really didn't do much for the auto-completion. It just prepared the dropdowns and such and those have changed so much that there's no point really. I think at least.

povrazor:
Okay. Good to know.

PR #1281, opened 2017-10-20
Fixes for refactoring mistake with piecharts

zwrawr:
I refactored the piechart code to use the correct spelling of labels in #1275, but didn't change it on the event stats page.

### DIFF
--- a/src/com/content-stats/stats-event.js
+++ b/src/com/content-stats/stats-event.js
@@ -94,7 +94,7 @@ export default class ContentStatsEvent extends Component {
 						'Jam',
 						'Compo'
 					];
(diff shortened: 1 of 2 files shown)


src/com/content-stats/stats-event.js
```diff
@@ -94,7 +94,7 @@ export default class ContentStatsEvent extends Component {
 						'Jam',
 						'Compo'
 					];
-					Data.push(<PieChart values={entries} lables={entrienames}></PieChart>);
+					Data.push(<PieChart values={entries} labels={entrienames}></PieChart>);
```
local-minimum: For reference, this could be written:

```
Data.push(<PieChart values={entries} labels={entrienames} />);
```

PR #1286, opened 2017-10-22
Layout Components

Noojuno:
Adds a few flexbox helper components (`Grid`, `Component`, `Row`, `Col`, `SelectableGrid`)  to make creating layouts easier. They'll be expanded on later when features need to be added but for now they generally work as needed.

Also converts the games listing to use a more generic `Grid` component for the sake of re-usability.


sgstair:
`commits.iso.tsv` file looks unintended.

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
 //        props.class.push("content-item-boxes");
 
 		var LoadMore = null;
-		const Games = [];
+		var Games = [];
```
local-minimum: I would use `let` rather than `var` so scope of variables is easier to read and predict

src/com/layout/col/col.js
```diff
@@ -0,0 +1,21 @@
...
+	constructor( props ) {
+		super(props);
+	}
+
+	render( {children, flex = 1, flexGrow, flexShrink, flexBasis = 100} ) {
+
+    var props = Object.assign({}, this.props, {flex: undefined, flexBasis: undefined, flexGrow: undefined, flexShrink: undefined});
+
+    var flexGrow = flexGrow != undefined ? flexGrow : flex;
+    var flexShrink = flexShrink != undefined ? flexShrink : flex;
```
local-minimum: Indentation and `var`

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
...
-				}
-			});
-
-			/*
-				As long as the number of items in the Games array
-				doesn't evenly divide by the number of columns
-				keep adding placeholder elements so that the last
-				row looks nice
-			*/
-			while ( Games.length % layout !== 0 ) {
```
local-minimum: Where is this handled downstream? I could not find it and as far as I have seen adding empty placeholders is needed to make the last row items keep the expected size if they are fewer than the number of columns. At least to make it so in all browsers. Should be needed for all flexers really.

src/com/layout/row/row.js
```diff
@@ -0,0 +1,15 @@
...
+export default class Row extends Component {
+	constructor( props ) {
+		super(props);
+	}
+
+	render( props ) {
+		return (
+			<div {...props} class="-row">
+        {props.children}
+      </div>
```
local-minimum: Indentation looks weird

src/com/layout/grid/selectable-grid.js
```diff
@@ -0,0 +1,31 @@
...
+import Grid from 'com/layout/grid/grid';
+import GridSelector from 'com/layout/grid/grid-selector';
+
+export default class SelectableGrid extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+			'columns': 3
+		};
```
local-minimum: Indentation hard to folloe

Noojuno:
@sgstair you're right, that was unintended. I'll push a commit to remove that

src/com/content-games/games.js
Noojuno: It's not handled downstream. In my testing it displayed the same in all browsers without placeholders so I removed it.

src/com/layout/row/row.js
Noojuno: The indentation issues are weird, not indented like that on my local copy but pushed like that... I'll see if I can fix it.

src/com/content-games/games.js
Noojuno: Good call.

src/com/content-games/games.js
local-minimum: That is fair, but just to be clear we had that low frequency ugly bug with last row sometimes being large, so I wouldn't want it reintroduced. If code can be made simpler though, then that isa strong argument for removing.

local-minimum:
How would you say this PR relates to my https://github.com/ludumdare/ludumdare/pull/1279

Does it make mine not so needed? Just thinking to minimize uneccesary work on Mike's part.

Noojuno:
@local-minimum You could build a shelf out of a grid pretty easily. I'm not sure really, it depends on the specific shelf implementation honestly.

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
 //        props.class.push("content-item-boxes");
 
 		var LoadMore = null;
-		const Games = [];
+		let Games = [];
```
local-minimum: Actually isn't games only used once below, better remove here then and define it on that row as a `const`

povrazor:
This sounds cool. I'm going to come back to this shortly (I'm going through non-merge issues right now).

src/com/layout/row/row.js
```diff
@@ -0,0 +1,15 @@
+import {h, Component} 				from 'preact/preact';
+
+export default class Row extends Component {
+	constructor( props ) {
+		super(props);
+	}
+
+	render( props ) {
+		return (
+			<div {...props} class={cN([props.class, "-row"])}>
```
povrazor: can be simplified to `cn(props.class, "-row")`

povrazor: Not sure exactly what this compiling to, given that `{...props}` will also have a `props.class`.

src/com/layout/grid/selectable-grid.js
povrazor: Yeah, it's minor, but it could be corrected.

src/com/layout/container/container.js
```diff
@@ -0,0 +1,15 @@
+import {h, Component} 				from 'preact/preact';
+
+export default class Container extends Component {
+	constructor( props ) {
+		super(props);
+	}
+
+	render( props ) {
+		return (
+			<div {...props} class={["container", props.class]}>
```
povrazor: `["container", props.class]` should be `cN(props.class, "container")` (`[]` is deprecated).

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
...
+    flexGrow = flexGrow != undefined
+      ? flexGrow
+      : flex;
+
+    flexShrink = flexShrink != undefined
+      ? flexShrink
+      : flex;
+
+    return (
+      <div class={cN([this.props.class, "-col"])} style={`flex: ${flexGrow} ${flexShrink} ${flexBasis}%; max-width: ${flexBasis}%;`} {...this.props}>
```
povrazor: Array brackets should be removed from `cN([this.props.class, "-col"])`. 

`this.props` should not be used in `render` (`props` and `state` are always available as arguments).

povrazor: Don't use backtick strings and `${}` syntax (not supported by IE).

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
+import {h, Component} from 'preact/preact';
+
+export default class Col extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render({children, flex = 1, flexGrow, flexShrink, flexBasis = 100}) {
```
povrazor: Prefer

```
render( props ) {
    let {flex = 1, flexGrow, flexShrink, flexBasis = 100} = props;
```

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
...
+      : flex;
+
+    flexShrink = flexShrink != undefined
+      ? flexShrink
+      : flex;
+
+    return (
+      <div class={cN([this.props.class, "-col"])} style={`flex: ${flexGrow} ${flexShrink} ${flexBasis}%; max-width: ${flexBasis}%;`} {...this.props}>
+        {children}
+      </div>
```
povrazor: `props.children`

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
...
+    super(props);
+  }
+
+  render({children, flex = 1, flexGrow, flexShrink, flexBasis = 100}) {
+    // let props = Object.assign({}, this.props, {
+    //   flex: undefined,
+    //   flexBasis: undefined,
+    //   flexGrow: undefined,
+    //   flexShrink: undefined
+    // });
```
povrazor: Values should never be initialized to `undefined`. Prefer `null`, as when we use `undefined`, you can't tell the difference between actual `undefined` values and user cleared values when debugging.

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
...
+
+    flexGrow = flexGrow != undefined
```
povrazor: To be clear, this test will pass if flexGrow is `undefined` or `null`. In this case, checking for `null` is preferred. 

If you explicitly want to test for `undefined` versus "not set", use `!==`. If it's inconsequential, a `!=` check is technically faster.

povrazor: This block of code is short and simple. Prefer a single line.

```js
flexGrow = (flexGrow != null) ? flexGrow : flex;
```

src/com/layout/col/col.js
```diff
@@ -0,0 +1,30 @@
...
+      ? flexGrow
+      : flex;
+
+    flexShrink = flexShrink != undefined
+      ? flexShrink
+      : flex;
+
```
povrazor: See above.

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
...
 		const {filter} = props;
 		if ( error ){
```
povrazor: `) {`

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
...
 			return <ContentError code="400">"Bad Request : Couldn't load games"</ContentError>;
 		}
 		else if( feed && feed.length > 0 )
```
povrazor: `else if ( feed && feed.length > 0 ) {`

src/com/content-games/games.js
povrazor: Yeah, this might need to be undone. I specifically asked @local-minimum to fix said bug.

src/com/content-games/games.js
```diff
@@ -9,10 +9,10 @@ import ContentItemBox					from 'com/content-item/item-box';
 import ContentCommonBody				from 'com/content-common/common-body';
 import ContentCommonBodyTitle			from 'com/content-common/common-body-title';
 
-import GridSelector						from './grid-selector';
-
 import ContentMore						from 'com/content-more/more';
 
+import SelectableGrid 							from 'com/layout/grid/selectable-grid';
```
povrazor: Prefer `CardGrid` (Cards being the new term for things currently known as `boxes`).

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
...
 
-			const gridClass = '-columns-' + layout;
+			Games = feed.map((r, index) => {
+				if ( ContentGames.matchesFilter(r.node, filter) ) {
+					return (<ContentItemBox
+						node={r.node}
+						user={props.user}
+						path={props.path}
+						noevent={props.noevent ? props.noevent : null} />
+					);
```
povrazor: This indentation can be improved.

```
					return (
						<ContentItemBox
							node={r.node}
							user={props.user}
							path={props.path}
							noevent={props.noevent ? props.noevent : null}
						/>
					);
```

src/com/content-games/games.js
```diff
@@ -166,55 +164,34 @@ export default class ContentGames extends Component {
...
-				As long as the number of items in the Games array
-				doesn't evenly divide by the number of columns
-				keep adding placeholder elements so that the last
-				row looks nice
-			*/
-			while ( Games.length % layout !== 0 ) {
-				Games.push(<ContentItemBox placeHolder={true} />);
-			}
-
 			if ( !props.nomore /*|| added >= 10*/ ){
```
povrazor: `) {`

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default class Grid extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render( {children, columns = 3} ) {
+    return (
+      <Container {...this.props} class={cN([this.props.class, "-grid"])}>
+        {children.map((child, index) => {
```
povrazor:
```js
props.children.map((child, index) => {
```

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+          return (
+            <Col flexGrow={0} flexBasis={100 / (columns)}>{child}</Col>
+          );
+        }, this)}
```
povrazor: remove tailing `}`

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render( {children, columns = 3} ) {
+    return (
+      <Container {...this.props} class={cN([this.props.class, "-grid"])}>
+        {children.map((child, index) => {
+          return (
+            <Col flexGrow={0} flexBasis={100 / (columns)}>{child}</Col>
```
povrazor: `100 / columns`

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+import Col from 'com/layout/col/col';
+import Row from 'com/layout/row/row';
+import Container from 'com/layout/container/container';
+
+export default class Grid extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render( {children, columns = 3} ) {
```
povrazor: prefer
```js
render( props ) {
    let {columns = 3} = props;
```

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+    return (
+      <Container {...this.props} class={cN([this.props.class, "-grid"])}>
```
povrazor:
```js
<Container {...props} class={cN(props.class, "-grid")}>
```

src/com/layout/grid/grid.less
```diff
@@ -0,0 +1,19 @@
...
+    & .-col {
+      padding: 0.5rem;
+      box-sizing: border-box;
+    }
+  }
+
+  &.-selectable {
+    padding-top: 0.75rem;
+  }
+}
```
povrazor: Indentation: Prefer tabs to two spaces (or 4 spaces and I wouldn't notice).

src/com/layout/grid/grid.less
```diff
@@ -0,0 +1,19 @@
...
+
+    & .-col {
+      padding: 0.5rem;
+      box-sizing: border-box;
+    }
+  }
+
+  &.-selectable {
+    padding-top: 0.75rem;
+  }
```
povrazor: Why would a container that can be selected have padding on top?

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
...
+        {children.map((child, index) => {
+          return (
+            <Col flexGrow={0} flexBasis={100 / (columns)}>{child}</Col>
+          );
+        }, this)}
+      </Container>
+    );
+  }
+}
```
povrazor: Indentation: Prefer tabs to two spaces (or 4 spaces and I wouldn't notice).

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,22 @@
+import {h, Component} from 'preact/preact';
+import Col from 'com/layout/col/col';
+import Row from 'com/layout/row/row';
+import Container from 'com/layout/container/container';
+
+export default class Grid extends Component {
+  constructor(props) {
```
povrazor:
```js
    constructor( props ) {
```

Spaces around arguments are used to differentiate function calls versus definitions.

src/com/layout/grid/selectable-grid.js
```diff
@@ -0,0 +1,31 @@
...
+      <div>
+        <GridSelector
+          defaultLayout={columns}
+          onChangeLayout={
+            (gridLayout) => {
+              this.setState({'columns': gridLayout});
+            }
+          }
+        />
+        <Grid {...props} columns={columns} class={cN([props.class, "-selectable"])}>
```
povrazor: Remove array brackets from inside `cN` call.

src/com/layout/row/row.js
Noojuno: JSX should (and in practice for this particular instance at least) be overridden by the property if it's set again. So it's basically just forcing `-row` to be there.

src/com/layout/grid/selectable-grid.js
Noojuno: This one really confuses me. In atom it looks fine, must be a weird indentation with spaces and tabs thing.

src/com/layout/grid/selectable-grid.js
```diff
@@ -0,0 +1,31 @@
...
+export default class SelectableGrid extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+			'columns': 3
+		};
+  }
+
+  render(props, {columns}) {
```
povrazor: Prefer
```js
render( props, state ) {
    let {columns} = state;
```

src/com/layout/grid/selectable-grid.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+          }
+        />
+        <Grid {...props} columns={columns} class={cN([props.class, "-selectable"])}>
+          {props.children}
+        </Grid>
+      </div>
+    );
+  }
+}
```
povrazor: Indentation: Prefer tabs to two spaces (or 4 spaces and I wouldn't notice).

src/com/content-games/games.js
Noojuno: So would that be `SelectableCardGrid` then? Given that there's a grid component separately and this   one is just with the selector.

src/com/layout/row/row.js
povrazor: Ah shoot. I forgot to delete that comment. Clearly it's working. :wink:

src/com/content-games/games.js
Noojuno: As far as I'm aware it was just a bug because of `flex-grow` being set to 1, whereas in the way the grid is done now it's set to 0. If I can reproduce the bug I'll pop that back in but I really haven't been able to in my testing.

povrazor: Okay, if you can try out the different sizes (1 wide, 2 wide, 3 wide, etc) and everything is equally sized, then it's good.

src/com/content-games/games.js
povrazor: What makes it selectable?

src/com/layout/col/col.js
Noojuno: This is the syntax I actually prefer. I was just following the syntax of the games page itself for state. Will change it to that.

src/com/layout/col/col.js
Noojuno: I believe there was a reason for using `undefined` over `null`. The purpose of this code was to avoid showing some useless preact props in the DOM itself and setting it to `null` would just show it as a `null` value rather than hiding the value.

Either way I decided that it didn't really matter and so I commented it out, when really I should've just removed the code itself.

src/com/content-games/games.js
povrazor: Equally sized in varying sized browser I should say.

src/com/layout/col/col.js
Noojuno: I was pretty sure buble handled the conversion of template strings to regular strings. I'll look into it and change it if that wasn't the case.

Admittedly I had issues getting the site actually working in anything lower than IE11 so I can't be certain, but it worked fine in IE11.

src/com/layout/grid/grid.js
Noojuno: This is a remnant from when it had to be `100 / (columns - 1)` because flexbox can be annoying. Will fix.

src/com/layout/col/col.js
povrazor: It's an unsafe transformation (meaning there are funny case that emit more code than expected), so I don't have it enabled.

src/com/layout/grid/grid.less
Noojuno: This is more an issue of actual naming than anything. I was struggling to think of another name for a component that had a `Grid` + a `GridSelector`.

Although I did just realise that it's meant to be only if it's both `-grid` and `-selectable` so I'll fix that.

povrazor: Cool. Yeah appears to be the core of the issue. What makes a `GridSelector` something that selects isn't exactly clear.

src/com/layout/col/col.js
povrazor: Yeah I only recently decided to change over (I think I've only discussed this with @local-minimum), so it's slowly being changed across the codebase.

src/com/layout/grid/selectable-grid.js
local-minimum: As a pythonite I growl every time I see tab in code. ;) probably a case of you having setting for how long a tab should be.

src/com/layout/col/col.js
povrazor: Fair enough. It gave me a good opportunity to mention `null` is preferred.

src/com/layout/col/col.js
Noojuno: Hmm okay damn. They're very useful to have.

I'll change them.

src/com/layout/grid/grid.js
povrazor: Makes sense. I'm assuming the code is more-or-less working properly now, so just catching minor things like this as I go.

src/com/layout/col/col.js
local-minimum: While I think `null` is preferable, here I think undefined is correct since the syntax asks for certain fields on `props` and they will be undefined not by setting that value but because they were actually undefined.

src/com/layout/col/col.js
```diff
@@ -0,0 +1,22 @@
...
+  render( props ) {
+    let {flex = 1, flexGrow, flexShrink, flexBasis = 100} = props;
+
+    flexGrow = (flexGrow != null) ? flexGrow : flex;
+    flexShrink = (flexShrink != null) ? flexShrink : flex;
+
+    let style = "flex: " + flexGrow + " " + flexShrink + "" + flexBasis + "%; max-width: " + flexBasis + "%;";
+
+    return (
+      <div class={cN(this.props.class, "-col")} style={style} {...this.props}>
```
povrazor: `...this.props` to `...props`

src/com/layout/col/col.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  render( props ) {
+    let {flex = 1, flexGrow, flexShrink, flexBasis = 100} = props;
+
+    flexGrow = (flexGrow != null) ? flexGrow : flex;
+    flexShrink = (flexShrink != null) ? flexShrink : flex;
+
+    let style = "flex: " + flexGrow + " " + flexShrink + "" + flexBasis + "%; max-width: " + flexBasis + "%;";
+
```
povrazor: Does this style stuff work when you resize the window?

src/com/layout/container/container.less
```diff
@@ -0,0 +1,13 @@
...
+
+  & .row {
+    flex: 1;
+    flex-direction: row;
+  }
+
+  & .col {
+    flex: 1;
+  }
+}
```
povrazor: In some code it's `.-row` and others it's `.row`.

src/com/layout/grid/grid-selector.js
```diff
@@ -0,0 +1,103 @@
...
+    );
+
+    if (expanded) {
+      const options = [
+        [
+          1, '1 per line', < div onclick = {
+            () => this.onSelectLayout(1)
+          }
+          class = "-click-catcher" />
+        ],
```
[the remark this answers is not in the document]
povrazor: This code is short enough to fit on one 80 character wide line though, so that might be a nicer way to structure ti

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,24 @@
...
+    super(props);
+  }
+
+  render( props ) {
+    let {columns = 3} = props;
+
+    return (
+      <LayoutContainer {...props} class={cN(props.class, "-grid")}>
+        {props.children.map((child, index) => {
+          return (
```
povrazor: Prefer
```
{
    props.children.map((child, index) => {
```

src/com/layout/grid/grid.js
```diff
@@ -0,0 +1,24 @@
...
+            <Col flexGrow={0} flexBasis={100 / columns}>{child}</Col>
+          );
+        })}
```
povrazor: Prefer
```
  });
}
```
Looks like the linter doesn't catch this.

src/com/layout/col/col.js
Noojuno: It sure does, it's all based on percentages.

src/com/layout/grid/grid-selector.js
```diff
@@ -0,0 +1,103 @@
...
+  render( props, {expanded, selected} ) {
+    let ShowDropDown = null;
+
+    const ShowToggle = (
+      <ButtonBase onclick={this.onToggleDropDown}>
+        <SVGIcon>cog</SVGIcon>
+      </ButtonBase>
+    );
+
+    if (expanded) {
```
povrazor: `if ( expanded ) {`

src/com/layout/col/col.js
povrazor: Cool

src/com/content-games/games.js
```diff
@@ -159,62 +157,43 @@ export default class ContentGames extends Component {
...
 
 		var LoadMore = null;
-		const Games = [];
 		const {filter} = props;
-		if ( error ){
+		if ( error ) {
 			return <ContentError code="400">"Bad Request : Couldn't load games"</ContentError>;
 		}
 		else if( feed && feed.length > 0 )
 		{
```
povrazor:
```js
else if ( feed && (feed.length > 0) ) {
```

Sorry, it's not your code, but I just saw this.

src/com/layout/col/col.js
```diff
@@ -0,0 +1,22 @@
+import {h, Component} from 'preact/preact';
+
+export default class LayoutCol extends Component {
+  constructor(props) {
```
povrazor:
```js
constructor( props ) {
```

src/com/layout/container/container.js
```diff
@@ -0,0 +1,15 @@
+import {h, Component} from 'preact/preact';
+
+export default class LayoutContainer extends Component {
+  constructor(props) {
```
povrazor:
```js
constructor( props ) {
```

src/com/layout/container/container.js
```diff
@@ -0,0 +1,15 @@
...
+    super(props);
+  }
+
+  render(props) {
```
povrazor:
```
render( props ) {
```

src/com/layout/grid/changeable-grid.js
```diff
@@ -0,0 +1,33 @@
+import {h, Component} from 'preact/preact';
+import LayoutGrid from 'com/layout/grid/grid';
+import GridSelector from 'com/layout/grid/grid-selector';
+
+export default class LayoutChangeableGrid extends Component {
+  constructor(props) {
```
povrazor:
```
constructor( props ) {
```

PR #1290, opened 2017-10-24
Removed extra {

Noojuno:


src/com/layout/grid/grid.js
```diff
@@ -18,7 +18,7 @@ export default class LayoutGrid extends Component {
               return (
                 <LayoutCol flexGrow={0} flexBasis={100 / columns}>{child}</LayoutCol>
               );
-          });
```
zwrawr: Do you need to remove this semi-colon ( ` props.children.map((child, index) => { [});` )

PR #1294, opened 2017-11-01
Barcharts!

zwrawr:
image

* Adds barcharts
* Pulls out the legend code from the piechart class so it's more generic and can be reused


src/com/visualization/barchart/bar.js
```diff
@@ -0,0 +1,42 @@
...
+    render( props ) {
+
+        // it's valid for height/width/index to be zero so we have to check it against undefined.
+        if ( !(
+			props &&
+			props['height']	!= undefined &&
+			props['index']	!= undefined &&
+			props['width']	!= undefined &&
+			props['color']
+		) ) {
```
povrazor: Prefer
```js
if ( !(props && (props.height != null) && (props.index != null) && (props.width != null) && (props.color != null)) ) {
```
or
```js
if ( !(
    props && 
    (props.height != null) &&
    (props.index != null) && 
    (props.width != null) && 
    (props.color != null)
) ) {
```
Noting that:
* The `!=` operator actually checks for both `undefined` and `null`, and `null` is preferred in this case.
* I'm guessing the `props.color` check also needs to be a `!= null` test

Alternatively, if a non-zero/non-empty-string is also a worthwhile thing to be testing for, consider:
```js
if ( !(props && props.height && props.index && props.width && props.color) ) {
```
Though without knowing the code, I could be wrong and `props.index` may be something where `0` should be legal.

src/com/visualization/barchart/bar.js
```diff
@@ -0,0 +1,42 @@
...
+            console.warn('Bar was created with invalid props', props);
+            return;
+        }
+
+        let height = props.height;
+        let width = props.width;
+        let index = props.index;
+		let color = props.color;
```
povrazor: Prefer
```js
let {height, width, index, color} = props;
```

src/com/visualization/barchart/barchart.less
```diff
@@ -0,0 +1,15 @@
...
+		align-self: center;
+
+		& > .-svg {
+			& > .-bar {
+				// maybe animate the bar's changing
+			}
+		}
+	}
+}
+
```
povrazor: last line should be a newline (not a space then newline)

src/com/visualization/legend/legend.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    render( props ) {
+
+        if ( !(
+			props &&
+			props['colors'] &&
+			props['names']
+		) ) {
+            console.warn('Legend was created with invalid props', props);
```
povrazor: prefer
```js
if ( !(props && props.colors && props.names) ) {
```
(notably `props.colors` vs `props['colors']`)

src/com/visualization/legend/legend.js
```diff
@@ -0,0 +1,48 @@
...
+            return;
+        }
+
+		let colors = props.colors;
+		let names = props.names;
+
```
povrazor: Prefer
```
let {colors, names} = props;
```

src/com/visualization/barchart/bar.js
zwrawr: 0 is legal for index height and width but not for color.
I keep forgetting that you don't need to use the `props.['key']` syntax

src/com/visualization/barchart/barchart.js
```diff
@@ -0,0 +1,63 @@
...
+
+export default class BarChart extends Component {
+
+	constructor( props ) {
+		super(props);
+
+    }
+
+
+    scaleValues ( values ) {
```
povrazor:
```
scaleValues( values ) {
```
No space between function name and bracket in a declaration

src/com/visualization/legend/legend.less
```diff
@@ -0,0 +1,29 @@
...
+	& > .-legend {
+		flex: 1;
+		padding-left: 15px;
+		padding-right: 15px;
+		align-self: center;
+		
+		& li {
+			list-style: none;
+			
+			& .-shape-circle {
```
povrazor:
```
& > .-shape-circle {
```

src/com/visualization/legend/legend.less
```diff
@@ -0,0 +1,29 @@
+
+.chart {
+	& > .-legend {
+		flex: 1;
+		padding-left: 15px;
+		padding-right: 15px;
+		align-self: center;
+		
+		& li {
```
povrazor:
```
& > li {
```

Just a few `>`'s, if we're sure they are always children of the prior element.

src/com/visualization/legend/legend.less
```diff
@@ -0,0 +1,29 @@
...
+			
+			& .-shape-circle {
+				display: inline-block;
+				vertical-align: middle;
+				width: 32px;
+				height: 32px;
+				border-radius: 50%;
+			}
+
+			& p {
```
povrazor:
```
& > p {
```

src/com/visualization/piechart/piesegment.js
```diff
@@ -10,7 +10,7 @@ export default class PieSegment extends Component {
     render( props ) {
 
         // it's valid for offset/angle to be zero so we have to check it against undefined.
-        if ( !(props && props['angle'] != undefined && props['offset'] != undefined && props['color']) ) {
+        if ( !(props && props.angle != undefined && props.offset != undefined && props.color) ) {
             console.warn('PieSegment was created with invalid props', props);
```
povrazor: - prefer `!= null` to `!= undefined`.
- wrap any check between `&&` and `||` that deals with multiple items in pair of brackets 
(i.e. `(props.angle != null)`, but not `props` or `props.color`).

Issue #1287, opened 2017-10-22
API realtime performance statistics


sgstair:
I've been thinking about a low-overhead mechanism for monitoring performance of the API on the live site in realtime, to allow us to generate some graphs and clearly see what exactly is going on and identify bottlenecks as the site is being used.

Here's the rough overview:
* Core parameters
  * This system will track usage in a fixed period - when the period expires it advances to the next period and stores the period summary data in the database.
  * The system stores data for the period in per-worker long-timeout APCU structures to avoid APCU locking, this requires some knowledge of the underlying hosting platform.
* API changes:
  * We will need to add some mechanism to identify the API name / identifier portion of the URL
  * When an API call completes, fetch the current worker + current period APCU structure, advance the counters for this API call
  * We will have, per API, a total count, total time, min/max time, and a minimal histogram of the values in that period.
  * When we enter a new reporting period, the new APCU key will be based on the period and will no longer accumulate to the old period.
* Magic changes
  * Magic task will add a quick check - If we have switched to a new reporting period. If so:
    * Fetch all of the available APCU structures for the previous reporting period, 
    * Merge and summarize the data, and store it into the database. 
      * Will store, per important API path: count, min, max, average, median, 99%, 95%, 90% percentiles.
    * Then, clean up the old unnecessary APCU nodes.
* API will have two new calls to interact with this data
  * Fetch database historical data- for providing an overview of how the API is performing over time
  * Get detailed data from current period (which will enumerate and summarize the data as done in the magic script, but provide more detailed output suitable for carefully examining what is going on in the API at the moment. (provide full histogram data)
* Site page to graph the historical and current data, so we can see what is going on.

povrazor:
> We will need to add some mechanism to identify the API name / identifier portion of the URL

Cool, okay. I might have a solution to this. I was already gung-ho about the idea of changing the structure of the API calls from a switch statement to a more flexible array type.

```js
 doAPI([
[ "node/get", API_GET, function( name ) {
  // Get a node
}],
[ "node/add", API_POST|API_AUTH, function( name ) {
  // create a node
}],
...
]);
```

Where the 1st argument is the name, the 2nd argument a list of common flags/restrictions we put on the call, and the last a function that actually contains the body of work to be done.

I hadn't decided how much name would be included in the 1st argument, but if there's reason to want the entire API endpoint, then so be it. That will make finding them with search much easier than they are now (i.e. right now there are sub-switch statements... would be awesome to just search for `node/get`).

[...]

Issue #1261, opened 2017-10-10
Adding game links require tags.


local-minimum:
This should not be, especially not thorough silent failing like it does now.

Note: does not direct apply to live site since it has tags.

sgstair:
Yeah, we can fix it by adding default tags in the simulator for now.
Long term, we should probably have a deploy script, that does a bunch of default setup actions - but that's presumably after we have a proper admin interface.

povrazor:
Alternatively we could fetch the tag list from the live server, and just re-create them locally.

local-minimum:
And I'd like to see the preact side deal with issues a bit nicer too

PR #1297, opened 2017-11-01
Rate Limiting + API Framework

### LINKED ISSUE #1285: API rate limiting
Here's the plan I'm building for rate limiting, based on conversations on the gitter:
* Rate limiting user token is based on the user (for authenticated API access), or the IP address, for unauthenticated access
* Charge is tracked via apcu_inc (and will apcu_add to create an entry on failure)
* All API access will charge one unit by default. Available API charge and maybe reset time will be added as a response header.
* Will define API rate limits in constants, probably:
  * Authenticated: 5000 API calls per 5 minutes for authenticated users. (well in excess of what should be necessary)
  * Unauthenticated: 1000 API calls per 5 minutes
* Will add API calls to allow charging more than one unit for expensive operations.
* Will add alternate charge pools for operations we want to throttle heavily without impacting the performance of the website (e.g. random game API).

sgstair:
* Resolve #1285 - Add API that allows for rate limiting via `rateLimit_Charge(<cost>)` which returns true if the API is allowed to proceed. Further you can charge to alternate pools via `rateLimit_Charge(<cost>,"poolname",<initial treasury>, <number of seconds before charge is reset>)` - The default pool is `default` with 5000 units to charge and resets after 5 minutes.
* Add a new API framework to produce the kind of "nice" flat API definitions that have been kicked around in various threads (recently, mostly on gitter and #1287, though I remember this being brought up in an earlier issue I don't recall)
* Notifications API has been ported to use this new API definition method here and is the first API to include rate limiting. Stats is also in progress in other pending work

Simulator changes:
* Address #1261 by having the simulator automatically create default tags if they're not present.
* Simulator now uses the username as the password for new users, padded with "a" to 8 characters if necessary.
* Some user creation refactoring, and now the simulator cycles users creating 5 new users per event.

Future:
* I'm tracking work to change the status code returned when an API is rate limited, I realize there are better codes available but haven't looked it up yet.


src/shrub/src/core/ratelimit.php
```diff
@@ -0,0 +1,84 @@
...
+	}
+	
+	// Allow the call to succeed if the pool was not depleted before the charge was made.
+	return ($newvalue + $charge) > 0;
+}
+
+/// Check how much charge remains in a rate limiting pool
+/// @param Pool to use, or nothing for the default pool
+/// @retval integer count of charge remaining, or null if there is no pool currently.
+function rateLimit_RemainingCharge($poolname = null) {
```
povrazor:
```php
function rateLimit_RemainingCharge( $poolname = null ) {
```

src/shrub/src/core/ratelimit.php
```diff
@@ -0,0 +1,84 @@
...
+	$value = apcu_fetch($key);
+	if ( $value === false ) {
+		$value = null;
+	}
+	
+	return $value;
+}
+
+
+function rateLimit_Key($poolname) {
```
povrazor:
```php
function rateLimit_Key( $poolname ) {
```

povrazor:
> About the only criticism I have is that the `API_CHARGE` part should maybe change in the future.
Rather than `API_CHARGE_1`, but rather we use constants for what class of API's are charged for the transaction.
> once we have them.
> That's a super easy change though.
> for now, `API_CHARGE_1` works as a placeholder.

> Probably do something clever, like `API_CHARGE_SEARCH | 1` to charge 1 unit of search to the call.
> or `API_COST_SEARCH`
> something like that

sgstair:
That is certainly something we can do in the future, yeah

PR #1301, opened 2017-11-02
Improve feedback when comment post fails

sgstair:
Previous behavior was to give no indication, just not add the comment + clear the comment field.

Tested:
* Can post comments normally
* Enable anonymous comments for node, refresh in another tab, disable anonymous comments, try to post anonymously, confirmed error message appears
* Enable chrome "offline" mode, confirm error message appears on post
* Confirm error message disappears when subsequently posting a comment successfully

image

### DIFF
--- a/src/com/content-comments/comments-comment.js
+++ b/src/com/content-comments/comments-comment.js
@@ -134,6 +134,7 @@ export default class ContentCommentsComment extends Component {
 		var user = props.user;
 		var comment = props.comment;
 		var author = props.author;
+		var error = props.error;
 
 		if ( author || comment.author == 0 ) {
 			var Name = "Anonymous";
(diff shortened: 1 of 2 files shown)


zwrawr:
I dislike the use of the word `node`. The vast majority of users have no idea what a node is (in regards to the site). 
Maybe thread, page or item.

sgstair:
eh, it's a raw message from the API, it makes sense in that context. Granted though, I don't really know how best to resolve this; Does the API error reporting defer to site UX concerns, or API UX concerns?

povrazor:
Well ideally that "Publish Anonymously" button should probably not be there in the first place on nodes where it's not allowed.

sgstair:
It isn't there on threads where it isn't allowed, but if you disable it before redrawing the page, it will not disappear - It was just an edge case I was thinking about, that's why I added the error codes.

sgstair:
(That and when I actually tried it, the site threw an exception, that's the real reason I spent some time on it. Then I realized there was no error reporting at all)

povrazor:
By the way, I don't know if this is the case, but Anonymous posting should be opt-in (defaulting to off), and at the moment only really show up on game pages. I really don't want to see anonymous comments on posts.

sgstair:
That is how things are, The button to allow anonymous comments only shows up when editing games.

povrazor:
Okay, so the UX issue is only when games get out-of-sync with their "can post anonymous comments" setting.

sgstair:
One other side note about my rationale for just emitting the raw error text rather than taking the time to format it nicely is: I don't really expect people to hit this case, but if they do, having a diagnosable error message is considerably better than the site just sitting there with no explanation, when pov inevitably gets an email from someone confused about why they can't post their comment.

povrazor:
Yeah, if folks should rarely see it (which I assume is the case), an exact error for me is preferred. :wink:

Issue #1300, opened 2017-11-02
Add CSS (List) Linting/Static Analysis


povrazor:
Less might support some static analysis features, but I'm not sure. 

The point is all our code should be tested by the toolchain.

PR #1305, opened 2017-11-06
Less Linting

zwrawr:
Addresses #1300 

Sets up Less linting as part of the tool-chain and fixes code style issues in existing less files. You will need to update your node modules using `npm install` to get stylelint installed.

### Important Changes
* .stylelintrc
* Makefile
* package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,22 +4,24 @@
   "description": "Custom CMS Platform",
   "main": "gulpfile.js",
   "dependencies": {
-    "buble": "Noojuno/buble",
-    "clean-css-cli": "*",
-    "es6": "*",
-    "es6-promise": "*",
-    "eslint": "*",
-    "less": "*",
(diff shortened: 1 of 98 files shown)


package.json
```diff
@@ -5,21 +5,21 @@
   "main": "gulpfile.js",
   "dependencies": {
     "buble": "Noojuno/buble",
+    "clean-css-cli": "*",
```
povrazor: I'm not sure I understand why the CSS cleaner was moved along side JS tools.

If I could document this file I would, but the blank lines (seen below) were separating tools needed by different concepts: JS, CSS, SVG.

zwrawr: Ah that change was unintentional I just ran `npm install stylelint --save`, and npm must of messed with the file.

zwrawr: Okay I've restored the formatting

PR #1323, opened 2017-11-10
Fix rendering error 

sgstair:
error occurs with when you have an @mention in post notification and also notified comments from that post, 
I didn't run into it on the simulator but it happened to me on the live site


zwrawr:
Also fixes #1325

PR #1367, opened 2017-11-15
WIP: Api stats

sgstair:
This addresses #1287. Presented here for awareness and testing purposes; I don't know how this is going to interact with the live site, but it's working pretty well for me in private. 
Feel free to merge if you don't feel it's too risky :)

Still to come:
* UX to render this data - Two pages, "API health" for general user consumption, and dev-focused per-API statistics page.
* Refactor API framework to switch to charge pool syntax  (e.g. API_CHARGE_SEARCH | 1)
* Switch all other APIs to using new API framework, so stats can discriminate between APIs
* Bit more characterization of how much time this adds to requests, and trying to mitigate that further.

What has been done:
* **ALL** API calls now record performance data. We keep track of, per API type (and all APIs together) - Total request count, Total time taken working on requests, and a histogram giving us detailed information on how long individual requests are taking (while still being cheap and not taking a lot of space)
* Stats API calls added to provide realtime detailed information on the current in-cache data, and historical data which is recorded to the database. The histogram is not persisted to the database, instead it is summarized to a set of percentile figures.
  * `/vx/stats/api` - Endpoint to get current realtime stats from the current period (up to 15 minutes). This fetches data from the redis structures and performs some basic computation, and it includes the full histogram data which gives a very accurate picture of the distribution of request completion times.
  * `/vx/stats/api/history` - Endpoint to get historical data. At the moment this is hardcoded to return the last ~48 records (12 hours) of data from the database. Further refinement to this API is anticipated.
* Magic cron task updated to flush previous period data to database when a new period is entered. This is expected to take 0ms typically and about 10ms when performing a flush (every 15 minutes).

Some other details:
* For clarity, think of the percentile data as follows: Percentiles["95"] is a request time, which 95% of requests are <= that much time. So if I see Percentiles["95"] = 0.001165, I know that 95% of the site requests are taking less than 1.165ms to complete. 
* We should observe the site with these changes and define some standards. e.g. "the site is in good health if P[95] < 20ms

### DIFF
--- a/private/magic.php
+++ b/private/magic.php
@@ -241,7 +241,4 @@ if ( $featured_id ) {
 			$db->commit();
 		}
 	}
-	
-	// Invalidate cached object that selects random games based on smart ratings.
-	nodeRandomGames_InvalidateCache();
 }
--- a/private/magic.php
+++ b/private/magic.php
@@ -7,6 +7,7 @@ require_once __DIR__."/".SHRUB_PATH."/cron.php";
 require_once __DIR__."/".SHRUB_PATH."/node/node.php";
 require_once __DIR__."/".SHRUB_PATH."/note/note.php";
 require_once __DIR__."/".SHRUB_PATH."/grade/grade.php";
+require_once __DIR__."/".SHRUB_PATH."/perfstats/perfstats.php";
 
 // This is a CRON job that regularly updates magic
 const MAX_ITEMS_TO_ADD = 500;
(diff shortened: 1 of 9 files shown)


povrazor:
I'm putting this off for now since I still need to merge in my redis session stuff to the main repo (it's on the main server but I kept it out of the VM). I just want to be sure it plays nice.

mikekasprzak:
Sorry for being months late. I've made the change to the dairybox setup, so a fresh install should have redis correctly configured the same as the server (using it as the php session save handler).

PR #1376, opened 2017-11-17
Search

povrazor:


povrazor:
Oops. I wasn't supposed to click that. Ah well.

public-api/vx/feed.php
```diff
@@ -0,0 +1,14 @@
...
+require_once __DIR__."/../config.php";
+
+include_once __DIR__."/".CONFIG_PATH."config.php";
+require_once __DIR__."/".SHRUB_PATH."api.php";
+require_once __DIR__."/".SHRUB_PATH."node/node.php";
+
+api_Exec([
+["feed", API_GET, API_CHARGE_1, function(&$RESPONSE) {
+
+	$RESPONSE['test'] = true;
```
local-minimum: Looks like debut code left in

public-api/vx/search.php
```diff
@@ -0,0 +1,15 @@
...
+include_once __DIR__."/".CONFIG_PATH."config.php";
+require_once __DIR__."/".SHRUB_PATH."api.php";
+require_once __DIR__."/".SHRUB_PATH."node/node.php";
+require_once __DIR__."/".SHRUB_PATH."/core/db_sphinx.php";
+
+api_Exec([
+["search", API_GET, API_CHARGE_1, function(&$RESPONSE) {
+	// q=what (like google)
+	
+	$RESPONSE['test'] = true;
```
local-minimum: Idem

public-api/vx/feed.php
local-minimum: *debug

povrazor: This is intended, to confirm the endpoint works.

PR #1484, opened 2017-12-06
Icons for notifications

zwrawr:
screenshot from 2017-12-06 04-50-22
screenshot from 2017-12-06 04-51-03
screenshot from 2017-12-06 04-51-16
screenshot from 2017-12-06 04-51-24

As discussed in #1449

PR #1493, opened 2017-12-07
change icons for notifications on your posts

zwrawr:
As mentioned at the of #1484

screenshot from 2017-12-07 09-25-38

Essentially if the notification has a `your` in it, the bubble is solid otherwise the bubble isn't solid.

### DIFF
--- a/src/com/content-notifications/notification.js
+++ b/src/com/content-notifications/notification.js
@@ -120,14 +120,14 @@ export default class NotificationItem extends Component {
 
 					return (
 						<NavLink {...navProps} >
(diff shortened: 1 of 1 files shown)


local-minimum:
ohno I did that too in my filtering thingy...

povrazor:
I believe #1494 does what this did (and much more). Double check the live site and make sure it's working as you wanted.

local-minimum:
Since we solved it differently I really don't want to swear that I didn't miss some case. But I did 3 fixes with the bubbles so the number of fixes matches.

zwrawr:
Your implimentaion does the same thing as this, just with the graphics opposite., closing

PR #1494, opened 2017-12-07
Notifications filtering

local-minimum:
image

Introduces a filter-bar with toggle-state buttons for what types of notifications to show in the notifications feed.

It also replaces icons so that mentions get their own icon (identical to #1493 ) but also fixes a missed case in the iconography. Further it changed up so that `bubbles` rather than `bubble` will be used for mult-notification where many people have caused the comment notification on the same thread.

The filter handles filtering unknown (other) notifications, but I feel updating to ensure that the notification is actually rendered is better left to a separate PR. It could simply use the `isNotificationOther` function and then do a generic render with the `question` svg as icon.

I would like a better icon for feedback class of notification than the current `bubble-empty` but I think that is also better for another PR


src/com/content-notifications/notification.js
```diff
@@ -1,9 +1,42 @@
 import {h, Component}	 				from 'preact/preact';
-
 import NavLink 							from 'com/nav-link/link';
-
 import SVGIcon 							from 'com/svg-icon/icon';
 
+
+export const isNotifactionType = (notification, notificationType) => {
```
sgstair: All of these - `Notification` rather than `Notifaction`

src/com/content-notifications/base.js
```diff
@@ -328,4 +339,46 @@ export default class NotificationsBase extends Component {
...
+		}
+		else if (comment !== false && isNotifactionComment(notification)) {
+			return true;
+		}
+		return other !== false && isNotificationOther(notification);
+	}
+
+	handleFilterChange(filterType, otherStuff) {
+		let oldFilter = Object.assign({}, this.state.filters);
+		if (filterType == 'comment') {
```
sgstair: I can't help but notice all these if statements could be replaced by a single line using oldFilter[filterType] = ...

local-minimum: That was my original idea, but I think the linter got really upset about it. I can give it a try and see if I'm remembering wrong.

povrazor: Feel free to share the line you want to make linter friendly. We can help.

src/com/content-notifications/notification.js
povrazor: notifaction-notifactions-everywhere

Sorry, I couldn't resist. I would recommend checking the entire changeset for `notifaction`. They're everwhere, not just here.

local-minimum: renamed all

cppchriscpp:
Just saw this in the ui and had to seek out the PR. I currently have 128 notifications from the past 2 days, and surely I have more to come, so THANK YOU!! 

I will say the ui was a little bit confusing the first time I got to the page, but I got through it. (And I sorely needed this feature)

Jezzamonn:
Oh... That's what the feature does.

Could we use a more checkbox style design? That's the UI standard for
on/off options

local-minimum:
Check boxes is a thing and having dropdown use same settings too in a later PR that Mike merged but hasn't deployed yet.

PR #1505, opened 2017-12-11
Notifications settings cookie

local-minimum:
Uses cookie not tied to the user in any way but shared for the computer for storing notifications settings. It makes communication between the navigation bar and the feed possible and keeping settings between sessions. Logic for this is in `shrub/js/notification/notifications.js` with the intent that if we move it to server-side settings, the components could stay as they are.

It adds a new notification type `feedback` for all notes on a post you're an author to. 

It only shows notification count for the types you're interested in.

It causes the notification count to be correct up to 40 new notifications given that all notifications you have are kept (not filtered out). If there are more than 40 new notifications, you'll get a number that may be lower than the actual number of notifications. It doesn't show that there may be more but code for doing such is included, but commented out.

If you get new notifications while the dropdown is expanded or you update your settings in the feed it will also update on next poll. 

If you mark notifications as read in the dropdown and there were more hidden notifications, they will show next poll. All filtered out notifications in the batch of 40 that were actually polled will be marked as read.

@sgstair Please check the changes in the php-code...

## Tests

* Make
* Run simulator several times
* Use the UI and behaves as expected.
* It can handle and render notifications of unknown type, but they will have very weird wording.


src/shrub/src/notification/notification_core.php
```diff
@@ -156,26 +156,25 @@ function notification_AddForPublishedNode( $node, $author, $type, $mentions = []
...
-			$users[] = $link['b'];
-		if ( $link['b'] == $node )
-			$users[] = $link['a'];
-	}
+	$authors = nodeMeta_GetAuthors($node_id);
+	$authors[] = $nodeauthor;
+	$authors = array_unique($authors);
+	
+	// Add users that are co-authors notifications
+	$users = array_merge($users, $authors);
```
sgstair: This is not necessary, as you are removing `$authors` from `$users` later.

src/shrub/src/notification/notification_core.php
```diff
@@ -195,19 +194,29 @@ function notification_AddForNote( $node, $note, $author, $mentions = [] ) {
 	$users = array_diff($users, $optout);
```
sgstair: You should also remove `$optout` users from your `$authors` list.

local-minimum:
@sgstair I updated the `php`-code with your comments, did some more structuring to it too.

sgstair:
@local-minimum Cool - updated PHP changes look good & correct. Thanks :)

povrazor:
Ah sorry. Please use `localstorage` instead of a cookie. Cookies inflate the size of HTTP requests.

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

src/com/ui/checkbox/checkbox.js
```diff
@@ -4,21 +4,14 @@ import UIButton							from '../button/button-div';	// NOTE: we don't need -link
...
-
 	render( props ) {
+		const iconName = (props.radio ? 'radio' : 'checkbox') + (props.value ? '-checked' : '-unchecked');
 		return (
-			<UIButton class={cN('ui-checkbox', props.class)}>
-				<UIIcon>{props.radio ? 'radio' : 'checkbox'}-{props.value ? 'checked' : 'unchecked'}</UIIcon>
+			<UIButton
+				class={cN('ui-checkbox', props.class)}
+				onclick={props.onclick}
+			>
```
povrazor: Please unbreak this out. It's not too long to fit on a line.

src/com/ui/checkbox/checkbox.js
```diff
@@ -4,21 +4,14 @@ import UIButton							from '../button/button-div';	// NOTE: we don't need -link
...
+				<UIIcon name={iconName}/>
```
povrazor: Please use a space before a `/>`.

PR #1510, opened 2017-12-19
Notification cleaning and more

local-minimum:
This was mostly intended to clean up the code of `notification.js` because it was really hard to work with and maintain.

Also fixed bugs: 

* not tracking friends and friends.
* correctly formatting how friends and people you follow are displayed (this was not rendering their names like it should have).
* fixed bug that would have anonymous feedback crash the notification-code.

Attempted to try and make old feedback-notifications that where typed as comments to still show and filter as feedback. However I would have to manually edit the database for some feedback notifications and make them classified as type `note`. But I don't have the skills for that. It seems at least not to break anything so merge should be safe even if we won't know if this fixed the thing or not until after.

### DIFF
--- a/src/com/content-notifications/base.js
+++ b/src/com/content-notifications/base.js
@@ -113,7 +113,7 @@ export default class NotificationsBase extends Component {
 
 		let soicialPromise = $Node.GetMy().then((response) => {
 
-			social.following = response.star ? response.star : [];
+			social.following = response.meta.star ? response.meta.star : [];
 			social.followers = response.refs.star ? response.refs.star : [];
 			social.friends = social.followers.filter((i) => social.following.indexOf(i) > -1);
 		});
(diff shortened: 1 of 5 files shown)


mikekasprzak:
> Attempted to try and make old feedback-notifications that where typed as comments to still show and filter as feedback. However I would have to manually edit the database for some feedback notifications and make them classified as type note. But I don't have the skills for that. It seems at least not to break anything so merge should be safe even if we won't know if this fixed the thing or not until after.

So just to be clear, worst case some notifications may have the wrong type assigned to them. They still display?

local-minimum:
Yes that's my estimate and if so, then it's equal to how it is on the live system now.


I've looked at about 100 - 200 notifications,  but it is hard to know if that covers all the cases.

PR #1530, opened 2017-12-22
Mygrades more stats

local-minimum:
Adds special graphs for two of the general sort orders:

* By grading order:
image
(Helps you see if you had a grumpy day or if you tend to give lower grades if you play too many games)

* By type of item:
image

### DIFF
--- a/src/com/page/event/mygrades/mygrades.js
+++ b/src/com/page/event/mygrades/mygrades.js
@@ -1,15 +1,12 @@
 import {h, Component} from 'preact/preact';
-
 import ContentSimple					from 'com/content-simple/simple';
-
 import ContentList				from 'com/content-list/list';
 import ContentLoading			from 'com/content-loading/loading';
-
 import UIButtonLink						from 'com/ui/button/button-link';
 import InputDropdown					from 'com/input-dropdown/dropdown';
 import SVGIcon 							from 'com/svg-icon/icon';
 import BarChart							from 'com/visualization/barchart/barchart';
-
+import PieChart							from 'com/visualization/piechart/piechart';
 import $Grade					from 'shrub/js/grade/grade';
 import $Node					from 'shrub/js/node/node';
 import $Comment					from 'shrub/js/comment/comment';
@@ -24,13 +21,11 @@ const pad = (number, digits) => {
 };
 
 export default class MyGrades extends Component {
-
 	constructor( props ) {
 		super(props);
 		this.state = {
 			'loading': true,
 		};
-
 		this.onSortByChange = this.onSortByChange.bind(this);
 	}
 
@@ -40,7 +35,6 @@ export default class MyGrades extends Component {
 			.then(r => {
 				this.setState({'grades': r.grade});
 			});
-
 		$Grade.GetMyList(this.props.node.id)
 			.then(r => {
 				this.setState({
@@ -53,7 +47,6 @@ export default class MyGrades extends Component {
 			.catch(r => {
 				this.setState({'error': r, 'gameIds': null, 'loading': false});
 			});
-
 		$Comment.GetMyListByParentNode(eventId).then(r => this.collectComments(r.comment));
 	}
 
@@ -79,7 +72,6 @@ export default class MyGrades extends Component {
 			}));
 			chunkStart += chunkSize;
 		}
-
 		return Promise.all(promises).then(r => {
 			this.setState({'myComments': nodesWithMyComments});
 		});
@@ -123,7 +115,6 @@ export default class MyGrades extends Component {
 			);
 			i += chunkSize;
 		}
-
 		return Promise.all(promises)
 			.then(r => {
 				this.setState({'loading': false, 'nodes': nodeMapping, 'gradeNames': gradeMapping});
@@ -136,7 +127,6 @@ export default class MyGrades extends Component {
 		const authorsMapping = new Map();
 		let promises = [];
 		const chunkSize = 20;
-
 		for (let i = 0; i < gameIds.length; i+=1) {
 			const node = mapping[gameIds[i]];
 			if (node.author && node.author > 0) {
@@ -151,7 +141,6 @@ export default class MyGrades extends Component {
 			}
 		}
 		authors = authors.filter((author, index) => authors.indexOf(author) == index);
-
 		let i = 0;
 		while (true) {
 			const chunk = authors.slice(i, i + chunkSize);
@@ -170,7 +159,6 @@ export default class MyGrades extends Component {
 			);
 			i += chunkSize;
 		}
-
 		return Promise.all(promises)
 			.then(r => {
 				this.setState({'authors': authorsMapping});
@@ -183,7 +171,6 @@ export default class MyGrades extends Component {
 		if (!authors) {
 			return [];
 		}
-
 		const itemAuthors = [];
 		const mainAuthor = authors[item.author];
 		if (mainAuthor) {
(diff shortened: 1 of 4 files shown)


local-minimum:
@mikekasprzak is there a particular issue with this PR? Just want to know so I can work on it.

mikekasprzak:
Ah no, sorry. Let me take a look.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -94,10 +86,10 @@ export default class MyGrades extends Component {
 		promises.push($Node.Get(eventId).then(r => {
 			const {meta} = r.node[0];
 			let gradeIndex = 1;
-			while (true) {
+			while ( true ) {
 				const gradeKey = `grade-${pad(gradeIndex, 2)}`;
```
mikekasprzak: Backtick strings should not be used.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -228,19 +215,19 @@ export default class MyGrades extends Component {
...
 							n += 1;
 							total += currentGrades[grade];
 						}
 						return [id, total / n];
 					})
-					.sort((a, b) => b[1] - a[1]) //Order by avg grade descending
+					.sort((a, b) => (b[1] - a[1])) //Order by avg grade descending
 					.map(elem => elem[0]); //Return ids
 			default: // Sorting by one of the grades
 				const gradeKey = `grade-${pad(sortBy - SORT_GRADE_AVERAGE, 2)}`;
```
mikekasprzak: Backtick strings should not be used.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -250,15 +237,18 @@ export default class MyGrades extends Component {
 		const gameIds = this.getSortedGames();
 		const shouldGradeNoGames = 20;
 		const hasResults = !loading && !error;
-		const ShowError = error ? <div class="-warning">Could not retrieve your votes. Are you logged in?</div> : null;
-
-		let ShowLoading = !gameIds && !error ? <ContentLoading /> : null;
+		const ShowError = error ? (<div class="-warning">Could not retrieve your votes. Are you logged in?</div>) : null;
+		const gradeKey = sortBy && ((sortBy > SORT_GRADE_AVERAGE) ? (`grade-${pad(sortBy - SORT_GRADE_AVERAGE, 2)}`) : null);
```
mikekasprzak: Backtick strings should not be used.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -442,53 +456,204 @@ class GradeStats extends Component {
...
 		const gradeLevels = [1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5, 5];
-		const ShowHistGraph = (
-			<BarChart
-				values={gradeLevels.map(v => hist[v] ? hist[v] : 0)}
-				labels={gradeLevels.map(v => `${v} star${v > 1 ? 's' : ''}`)}
-				use_percentages={false} />
+
+		if ( focusGrade ) {
+			votesData = this.getHistogram(grades, focusGrade);
+			DetailGraphTitle = `Distribution of stars (${gradeNames[focusGrade]}):`;
```
mikekasprzak: Backtick strings should not be used.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -442,53 +456,204 @@ class GradeStats extends Component {
...
+			DetailValues = gradeLevels.map(v => votesData[v] ? votesData[v] : 0);
+			DetailLabels = gradeLevels.map(v => `${v} star${(v > 1) ? 's' : ''}`);
```
mikekasprzak: Backtick strings should not be used.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -442,53 +456,204 @@ class GradeStats extends Component {
...
+			DetailGraphTitle = 'Total average grade per day';
+			DetailValues = binnedGradesDays.map(date => binnedGrades[date] ? this.averageArray(binnedGrades[date].map(e => e[1])) : NaN);
+			DetailLabels = binnedGradesDays;
+
+		}
+		else {
+			votesData = this.getGlobalHistogram(grades, gradeNames);
+			DetailGraphTitle = 'Total distribution of stars';
+			DetailValues = gradeLevels.map(v => votesData[v] ? votesData[v] : 0);
+			DetailLabels = gradeLevels.map(v => (`${v} star${(v > 1) ? 's' : ''}`));
```
mikekasprzak: Backtick strings should not be used.

src/shrub/src/core/db_mysql.php
```diff
@@ -729,3 +729,22 @@ function db_QueryFetchIdKeyValue( $query, ...$args ) {
...
+				$ret[$value[0]] = [];
+			}
+			
+			$ret[$value[0]][$value[1]] = $value[2];			
+			$ret[$value[0]][$value[1] . $metaName] = $value[3];			
+		}
+	}
+	
+	return $ret;
+}
```
mikekasprzak: There should not be table-specific functions inside `db`.

src/shrub/src/grade/grade_core.php
```diff
@@ -28,14 +28,16 @@ function grade_GetByNodeAuthor( $node_id, $author_id ) {
...
-			value
+			value,
+			timestamp
 		FROM ".SH_TABLE_PREFIX.SH_TABLE_GRADE."
 		WHERE author=? AND parent=?
 		;",
+		'-timestamp',
 		$author_id,
 		$parent_id
 	);
```
mikekasprzak: If you want different output, call a different function, and/or transform it here.

src/com/page/event/mygrades/mygrades.js
local-minimum: Why is that btw? It makes it so so much more readable.

src/shrub/src/grade/grade_core.php
mikekasprzak: Okay, not exactly what I meant (though technically correct). 

What I more meant was a different `db_QueryFetch` call should be used.

`db_QueryFetch` will give you an array of data something like this:

```js
[
    {
        'id': 54,
        'name': "geoffry",
        'value': 77,
        'timestamp': "2008-11-11T00:00:00Z"
    }
]
```

And `db_QueryFetchWithIntKey('id', ...)` will give you something like this:

```js
{
   '54': {
        'id': 54,
        'name': "geoffry",
        'value': 77,
        'timestamp': "2008-11-11T00:00:00Z"
    }
}
```

mikekasprzak:
Because it requires a specialty transform that Buble (compiler) classifies
as "unsafe". That and doing it without isn't difficult, nor necessarily
more verbose.

Consistency is extremely helpful when maintaining large codebases. Having
several different ways of building strings that aren't necessarily better
is more difficult to maintain than having fewer. For example, in places
where we need to string replace something that's being renamed (i.e. notes
to comments), you're less likely to miss cases because you forgot about the
other syntax. This is why there's an effort to move all class string
building to the cN function: so all class building code can be tracked down
easily, because it's consistent.

src/com/page/event/mygrades/mygrades.js
```diff
@@ -320,54 +330,53 @@ export default class MyGrades extends Component {
...
-				{ShowStats}
 				{ShowSorting}
+				{ShowStats}
 				{ShowResults}
 			</div>
         );
     }
 }
 
 class GradedItem extends Component {
```
mikekasprzak: Seperate file

src/com/page/event/mygrades/mygrades.js
```diff
@@ -413,82 +426,225 @@ class GradedItem extends Component {
 }
 
 class GradeStats extends Component {
```
mikekasprzak: Seperate file

PR #1535, opened 2017-12-23
User groups!

local-minimum:
This adds the ability to post news (if you have privileges for it), from new news feed.

ldmakenews

## Important!

The `sh_user` table needs updating. And I don't fully understand the `table_create.php` files so I left it out. It doesn't have to be `SMALLINT`  but I think it ought to be sufficient with 16 combinatorial categories.

```
ALTER TABLE `sh_user` ADD `user_group` SMALLINT UNSIGNED NOT NULL DEFAULT '0' AFTER `last_auth`;
```

## Things to consider

* Currently posting to current event is disabled because I don't know how to validate a particular node as current event on the end-point.
* I didn't dare mess with current db-calls so the user post is gotten more times than is strictly needed in some scenarios.
* The user group belonging is public
* The only way to put a user in the "can post general news"-group is to update the `user_group` value of that user in the database. To have general news rights (but no other group belongings) set value to 64 or you can use `userGroup_GetStatusCodeFromFlags` passing in the group constants that the user should have and it returns the correct value.

### DIFF
--- a/public-api/vx/node.php
+++ b/public-api/vx/node.php
@@ -512,7 +512,8 @@ switch ( $action ) {
 			$type = coreSlugify_Name(json_ArgShift());
 			$subtype = coreSlugify_Name(json_ArgShift());
 			$subsubtype = coreSlugify_Name(json_ArgShift());
-
+			$createNewsPrivileges = userGroup_GetUserHasStatus($user_id, USER_GROUP_FLAG_CAN_NEWS, USER_GROUP_FLAG_ACTIVE_EVENT_HOST);
+			
 			$fulltype = $type;
 			if ( $subtype )
 				$fulltype .= '/'.$subtype;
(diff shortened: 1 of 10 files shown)


mikekasprzak:
Sorry, I know a lot of work went in to this, but for security reasons the `sh_user` table should be inaccessible from the API. That may sound unusual, but you'll note that the user API never actually returns data, just results. You feed it things, and states do change, but nothing is ever directly returned.

Also, permissions to do things are handled by private (protected) user metadata. There are not user levels. Rather, specific users are given permissions for specific functions.

local-minimum:
Can you be more specific where and how that should handled? Is the problem getting any info from that table?  It would be rather easy to make separate user groups table instead if that. And no worries. Got more familiar with the php.

mikekasprzak:
The `sh_user` table is used for one thing only: authentication. Are you who you say you are? It's the only table in the database that contains sensitive information, and if gets compromised, that's bad for everyone. Eventually extreme measures may be taken, such that this table is entirely isolated from the rest of the data.

Like all data, Users are nodes. Anything special above-and-beyond the norm is implemented in nodes via metadata.

Every piece of metadata has a "scope". Different scope levels dictate who has access to it. **Public** scope (level 0) is the metadata that public information, and gets cached alongside the nodes. Then you have **Shared** scope, which is visible privately to the two nodes in question (i.e. `a` and `b`). Then you have **Protected** scope, which is similar to **Shared** scope, but only `a` knows about it. Finally, you have **Server** scope, a level of data that is ***never*** sent to the end user.

Server side, you have access a users non-public data. You just need to ask for it.

Client side, you have the `user` variable that gets passed down the chain. This not only contains a complete node structure, but it also contains a `private` section with their private metadata inside (everything **Shared** and **Protected**). Scope doesn't matter to the client. It's discarded server side before its passed down. 

(and on a sidenote, `src/com/content-notifications/base.js` should not be making a call to `$Node.GetMy`. Again, this data is available via `user.private.meta` and `user.private.refs`).

User permissions should be **Protected**. This means the user themselves knows about them, but nobody else.

Client side, you simply check `user.private.meta` for the specific permission. Something like `can-post-news`.

```js
if ( parseInt(user.private.meta['can-post-news']) ) {
   // User can post news
}
```

Server side it's a little trickier (because this hasn't been needed yet), but the general idea is that the code that checks if a piece of metadata can be set needs an extra layer, one that specifies the pre-requisite permission. If that pre-requisite passes, then the API lets you perform the action.

The pre-requisites system doesn't exist yet. Nobody has needed it. That's not to say it's not necessary, just things like news posts were few and far between, that the couple minutes it takes me to flick the switch manually didn't make it an immediate priority.

local-minimum:
I still have some questions but this is a great start.  But since it is of general and permanent interest, I'll start a wiki page for it. I hope you think that's good and will help correct my assumptions and guesses.

PR #1548, opened 2017-12-30
Autocomplete!

local-minimum:
ld_autocomplete4

This introduces auto-completion thingy for at-names and emojis.

When typing if you start a pattern that matches being an at-name the at-name-autocompleter activates. If you start a pattern of something being an emoji-code that autocompleter activates.

Autocompleters filter their list of known things (`window.emoji.emojiList` or the authors of the node and all the comments) for things that match what has been written. This does not have to be start of name/word. But needs to be exact match (including case).

Autocompleters score matches and shows the best at top.

User can change selected thing to autocomplete by using arrow keys (up / down).
User can select selected thing by pressing enter.

Because there's a lot of work to get overflow of the css not to limit the list extent and to figure out where the caret is in the textarea, I opted for using a fixed position.

Because it felt important that the autocomplete should disappear if you defocus the textarea I opted for not actually allowing mouse clicking on the options.

### DIFF
--- a/src/com/content-comments/comment-autocomplete.js
+++ b/src/com/content-comments/comment-autocomplete.js
@@ -4,30 +4,10 @@ import UIButton						from 'com/ui/button/button';
 class Autocompletions extends Component {
 	constructor(props) {
 		super(props);
-		this.onKeyUp = this.onKeyUp.bind(this);
 		this.onKeyDown = this.onKeyDown.bind(this);
-		//this.handleAbort = this.handleAbort.bind(this);
-		//this.updateSelected = this.updateSelected.bind(this);
-		//this.selectSelected = this.selectSelected.bind(this);
-		//this.removeCharacter = this.removeCharacter.bind(this);
-		//this.updateText = this.updateText.bind(this);
 	}
 
 	onKeyDown( e ) {
-		e.preventDefault();
-		switch (e.key) {
-			case "Shift":
-				this.setState({'shiftKey': true});
-				break;
-			default:
-				this.updateText(e.key, this.state.shiftKey);
-		}
-		return false;
-	}
-
-	onKeyUp( e ) {
-		//console.log("Key Event", this, e);
-		e.preventDefault();
 		switch (e.key) {
 			case "Escape":
 				this.handleAbort();
@@ -41,15 +21,8 @@ class Autocompletions extends Component {
 			case "Enter":
 				this.selectSelected();
 				return false;
-			case "Shift":
-				this.setState({'shiftKey': false});
-				return false;
-			case "Backspace":
-				this.removeCharacter();
-				return false;
-			default:
-				return false;
 		}
+		return true;
 	}
 
 	componentWillReceiveProps( nextProps ) {
@@ -91,49 +64,6 @@ class Autocompletions extends Component {
 		};
 	}
 
-	removeCharacter() {
-		const {text, cursorPos, matchStart, matchEnd} = this.state;
-		const {onSelect} = this.props;
-		const updatedText = text.slice(0, cursorPos - 1) + text.slice(cursorPos);
-		if (cursorPos - 1 == matchStart) {
-			if (onSelect) {
-				onSelect(updatedText, cursorPos - 1);
-			}
-		}
-		else {
-			const matchObj = this.getMatch(updatedText, cursorPos);
-			if (matchObj) {
-				this.setState({'text': updatedText, 'cursorPos': cursorPos - 1, 'match': matchObj.match, 'matchEnd': matchObj.matchEnd});	
-			} else {
-				//TODO: first characters..
-			}
-			
-		}
-	}
-
-	updateText( key, hasShift ) {
-		if (key.length == 1) {
-			const {text, cursorPos, matchEnd} = this.state;
-			const {onSelect} = this.props;
-			const updatedText = text.slice(0, cursorPos) + (hasShift ? key.toUpperCase() : key) + text.slice(cursorPos);
-			const matchObj = this.getMatch(updatedText, cursorPos);
-			if (matchObj) {
-				this.setState({'text': updatedText, 'cursorPos': cursorPos + 1, 'match': matchObj.match, 'matchEnd': matchObj.matchEnd});
-				if (this.getMatching(matchObj.match).length == 0) {
-					if (onSelect) {
-						onSelect(updatedText, cursorPos + 1);
-					}
-				}
-			}
-			else {
-				this.setState({'selected': null, 'text': updatedText});
-				if (onSelect) {
-					onSelect(updatedText, cursorPos + 1);
-				}
-			}
-		}
-	}
-
 	selectSelected() {
 		const {text, cursorPos, match, matchStart, matchEnd} = this.state;
 		const {onSelect} = this.props;
@@ -150,7 +80,6 @@ class Autocompletions extends Component {
 	}
 
 	updateSelected(indexChange) {
-
 	}
 
 	handleAbort() {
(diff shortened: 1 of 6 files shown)


mikekasprzak:
Ah, the placement of the popup is weird. It should be relative the cursor like it is here on GitHub when you type a `:`. 

It looks like they do it by setting `style` of the popup, and setting it to a hard-coded pair of coordinates (i.e. `style="top: 31px; left: 149.047px;"`). It's also always there, and set visible (`display: block;` vs `display: none`) as a user types.

mikekasprzak:
Great PR. I'm still playing, but you've taken care of most of my concerns (focus, partial matches, etc).

Another change I'd suggest is lowering the number of displayed items to `8`, to be friendlier to smaller screens/mobile.

mikekasprzak:
Oh I got another: for @name matching, right now it's case sensitive. It should be insensitive.

screenshot from 2018-01-02 19-52-02

In this example, if I was to type a lower case `v`, it wouldn't complete the text.

Pushing `TAB` to complete doesn't work.

After autocompleting, you should include a space (' ') after the inserted symbol/name, and adjust the cursor accordingly.

Also a bug: Pushing `ENTER` to autocomplete will sometimes trigger an error.

```
nav-event.es.js:46 Uncaught TypeError: Cannot read property 'substr' of null
    at AutocompleteEmojis.getOptions (nav-event.es.js:46)
    at AutocompleteEmojis.selectSelected (event.es.js:114)
    at AutocompleteEmojis.onKeyDown (event.es.js:53)
    at ContentCommentsComment.onKeyDown (stats-event.es.js:28)
    at InputTextarea.onKeyDown (games.es.js:18)
    at HTMLTextAreaElement.eventProxy (dropdown.es.js:25)
```

Reproducing is difficult. I think it's in part switching modes (emoji to @name), part focus loss, part browser tab focus loss. Honestly I'm not 100% sure how to repro, but switching and coming back will occasionally trigger that error.

ajayyy:
If mouse clicks don't autocomplete, then this will be unusable on mobile

local-minimum:
Now click works! In fact I believe I fixed all but the layout issue.

mikekasprzak:
Here's a short clip of me reproducing the bug in my VM.

https://www.youtube.com/watch?v=v9V6-EZWiKY

First I type `:cat` and push enter. Good.

Next I type `@ver` for a user `@VeraTutina` (generated by simluator). When I push enter, in this case the cursor re-positions to the start of the word and nothing happens.

**Other Side Effects**
* once this bug kicks in, it becomes impossible to use the enter key to insert newlines.
* pushing enter doesn't always have the same behaviour. Sometimes the text outright deletes. Other times it replaces with other random things.
* once the bug kicks in, typing a `:` no longer pops up the autocomplete popup immediately. you need to type another character, or delete it first.

https://www.youtube.com/watch?v=mctMWsB-lLk

Above, if I type a `:`, delete it, then type the `@`, push enter, I get `:x:`

My best guess is the wrong array is being used somewhere. Some array of data that in some rare cases produces the same result.

mikekasprzak:
Okay, I think this is the last thing.

https://www.youtube.com/watch?v=7JJLOHZygqI

To demonstrate, the current behaviour lets you walk over emoji with the arrow keys, and it switches you to emoji pick mode. This is a problem because:

* if you have a large document, it hinders your ability brows up and down the document using the arrow keys: you sometimes get stuck.
* If you use an emoji like `:cat:`, there are multiple emoji containing the term "cat", so pushing down (to navigate a document) selects another emoji.

Later in the video I demonstrate github's box, which simply treats it as text, _unless_ you make an edit to something inside a pair of `:emoji:` colons, or as part of an `@name`. Again, this is ideal, because it means the autocompletion doesn't hinder your ability to compose large posts, descriptions, and comments.

Finally, two more things:
* Pushing the **ESC** key should exit autocompletion edit mode. Right now it just makes the autocomplete box go away (though it might be browser doing something clever with focus).
* and... ah crud. I can't actually remember what this was. I'll add another comment when I do.

mikekasprzak:
Just posting this here so I don't lose the link:

https://github.com/component/textarea-caret-position/blob/master/index.js

mikekasprzak:
Also I mentioned another issue in Gitter, but there's been a bunch of noise, so let me post it again just you have a reference.

https://www.youtube.com/watch?v=YMkZVr6XzJw

In this video, it shows everything is working as expect, except if you type an `:emoji:` without using ENTER or TAB to autocomplete it. If you fully type out the command (including both colons), navigate to another emoji with the arrow keys, it thinks it's still in edit mode.

Not shown, if you push ENTER/TAB while on one of the autocompletions, it will end autocomplete mode, and return to the expected state.

(excuse the end of the video, I pushed the wrong record stop hotkey)

mikekasprzak:
The other issue is a bit more obscure.

https://www.youtube.com/watch?v=9h60ll0eBQE

If you make an edit to an `:emoji:`, don't finish, and navigate to other emoji it's fine. However if you navigate to an `@name`, it thinks its still in edit mode.

So to mitigate all these sort of issues, you might want to auto-end autocompletion edit mode any time the cursor moves somewhere outside a thing that can be autocompleted.

mikekasprzak:
And just so it doesn't get missed again, pushing the `ESC` key should exit autocompletion mode no matter what.

mikekasprzak:
One more. 

https://www.youtube.com/watch?v=QpwAIv2j7S8

When you navigate left/right inside a string, it changes the autocomplete string. It should not do this (if your word is bread, it should still be bread no matter where you are in the string).

This can be abused to trigger an insert glitch. Pushing Enter/Tab while the autocomplete thinks the string is shorter will insert some garbage (the unused rest of the string).

local-minimum:
Going to take notes here:

- [x] Manual completion of thing exits edit mode
- [x] `Esc` exits edit mode
- [x] Swapping back and forth in incomplete patterns or incomplete and complete patterns between at-names and emojis.
- [x] Navigating inside autocomplete changes matched segment

I hope I didn't miss anything.

mikekasprzak:
Yep! This is pretty solid. I'm having a hard time thinking of how to break it now.

mikekasprzak:
We're getting pretty obscure. :wink:

https://www.youtube.com/watch?v=tgd6W2Ykqvc

When you use the `ESC` key, it breaks all further autocompletes that begin with a colon `:`.

src/com/content-comments/comment-autocomplete.js
```diff
@@ -0,0 +1,303 @@
...
+			return;
+		}
+
+		this.executeSelect(item.name);
+	}
+
+	executeSelect( selected ) {
+		const {matchStart, matchEnd, match} = this.state;
+		const {onSelect, text} = this.props;
+		selected = `${selected} `;
```
mikekasprzak: Don't use backtick strings. Prefer `selected + ' '`.

mikekasprzak: or actually, in this case `selected += ' ';` should work.

local-minimum:
So think I fixed those two too now.

mikekasprzak:
I am so sorry. :wink:

https://www.youtube.com/watch?v=6BVxuwh14DM

Under certain circumstances, usually when a colon (`:`) is added before a string that should become an emoji (or should have been an emoji) it doesn't necessarily realize its related. Pushing right will let you add letters, like the earlier glitch, and pushing enter will do the insertion glitch.

mikekasprzak:
https://www.youtube.com/watch?v=FZYYcktRb_A

I can trigger an error if I delete the space and part of a side-by-side emoji.

Issue #1501, opened 2017-12-08
Support Firefox and Safari's Login/Password remember'er


povrazor:
Right now Firefox and Safari's feature that will auto-fill your login and password do not work. They work in Chrome (was automatic), but not Firefox/Safari.

local-minimum:
FF fills in mine, but it is glitchy.

mikekasprzak:
If you want to do some digging in to this, it would be appreciated. I'm not 100% sure why it even works in Chrome.

local-minimum:
I did some first checking. For some reason I have one user name and password combo saved in Firefox and using them I noticed that when selecting the saved one, in most manners don't trigger an onchange of the username field. It I type the first letter and with arrows and tab select the saved credentials then it actually works. So my hope is that some event may fire. I just have to find it. Else we have to ref the inputs and grab their values that way.

PR #1550, opened 2017-12-31
Remember password

JonesAndrew:
https://github.com/ludumdare/ludumdare/issues/1501

The diff is a little messed up, but all I did was put the inputs into a form. I'm pretty sure this is mandatory for safari and firefox to recognize the fields.

### DIFF
--- a/src/com/dialog/login/login.js
+++ b/src/com/dialog/login/login.js
@@ -71,30 +71,32 @@ export default class DialogLogin extends Component {
 		}
 
 		return (
-			<DialogCommon ok oktext="Log In" onok={this.doLogin} cancel {...new_props}>
-				<div>
-					<div class="-input-container">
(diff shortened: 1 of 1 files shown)


local-minimum:
lgtm, and welcome to the dev-team :) Did you try it out locally too with setting up vagrant and such?

JonesAndrew:
Thanks. Ya, I followed all the steps and set up the vagrant. I was able to test out safari and that looked good, but I always forget how you trick firefox into letting you save passwords over HTTP.

mikekasprzak:
Okay, I've pushed the change to the live server. Can someone with Safari confirm it's working? I've confirmed with Chrome that it didn't break anything.

As for Firefox, I might have a solution a bit later. Firefox for me complains that it doesn't want to offer to save passwords because of HTTP. That said, I might actually be able to get HTTPS working locally, it's just a little wonky to set up.

JonesAndrew:
It worked for me on safari, but not on firefox.

mikekasprzak:
Ah well, that's still an improvement. Thanks for the change. :)

PR #1551, opened 2018-01-02
Made simulator put events under correct node

local-minimum:
Before the events didn't end up under the `events` node.

Fixed it.

Tested by running simulator and making sure `/events` listed the event.

Seems fine.

### DIFF
--- a/sandbox/simulator/main
+++ b/sandbox/simulator/main
@@ -100,6 +100,17 @@ define("TOP_STATS",10); // keep top 10 stats
 
 // Figure out event index to use for this run, and start loggers.
 
+$events_root = db_QueryFetch(
+		"SELECT
+			n.id,
+			".DB_FIELD_DATE('n.modified', 'modified')."
+		FROM
+			".SH_TABLE_PREFIX.SH_TABLE_NODE." AS n
+		WHERE n.type = 'events'
+		LIMIT 1
+		;"
+	);
+	
 $events = db_QueryFetch(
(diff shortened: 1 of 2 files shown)


sgstair:
Looks like it should work without issues. Thanks!

sandbox/simulator/main
```diff
@@ -100,6 +100,17 @@ define("TOP_STATS",10); // keep top 10 stats
 
 // Figure out event index to use for this run, and start loggers.
 
+$events_root = db_QueryFetch(
+		"SELECT
+			n.id,
+			".DB_FIELD_DATE('n.modified', 'modified')."
```
mikekasprzak: Date is unused.

sandbox/simulator/main
```diff
@@ -113,6 +124,7 @@ $events = db_QueryFetch(
 	);
 
 $eventindex = count($events) + 1;
+$eventsId = $events_root[0]['id'];
```
mikekasprzak: Modified date is unused, you use a LIMIT 1 in the query, and you only care about a single result value. In this case, you can use `db_QueryFetchValue`, and store the returned value in `$eventsId` directly.

sandbox/simulator/main
local-minimum: fixed.

sandbox/simulator/main
local-minimum: fixed

PR #1557, opened 2018-01-04
Stops travis failing

zwrawr:
You've enabled travis but npm test function fails by default. This should stop travis failing till we add real tests.

#1548


mikekasprzak:
Yeah, it came up in discussion last night, so by request I flicked the switch to see what would happen.

PR #1566, opened 2018-01-17
User management shrub tool

local-minimum:
Start of a user management shrub tool.

Allows toggling and listing status of user's rights.
At the moment can only toggle create-event right.


mikekasprzak:
Accepting because it doesn't conflict with anything.

Issue #1307, opened 2017-11-07
Duplicate theme suggestions submission


Stuntddude:
I clicked on the "Submit" button twice for my third theme suggestion and managed to submit it again as a fourth suggestion. When I refresh the page (or leave and come back), it's still there. Looks like this:

<img width="546" alt="screen shot 2017-11-07 at 9 19 32 am" src="https://user-images.githubusercontent.com/6372134/32501671-3817ff42-c39e-11e7-99bd-32f259f9be9e.png">

povrazor:
To be clear the bug here is that the number of suggestions can be pushed beyond 3 by brute forcing it. If someone wants to submit the exact same theme 3 times, that is their choice, though we should warn them with a :warning: that we do remove duplicates.

povrazor:
Oh, and the fix would be to wrap this in a transaction.

http://php.net/manual/en/mysqli.begin-transaction.php

mikekasprzak:
Still needs the transaction stuff.

mikekasprzak:
It's still a bug, and frankly it's an easy fix (wrap it and a check in DB
transaction statements). No more bugs. 😉

PR #1569, opened 2018-01-19
Theme ideas

local-minimum:
ld-themes-ideas2

* Cleaning up the code
* Making it look more like the rest of the site
* Making it harder to dual post #1307 
* Making sure the number of suggestions left will always be sensible.
* Making your suggestions stand out a bit and not just look like ordinary text.
* Showing error messages and making sure those are understandable to the user.


mikekasprzak:
Minor suggestion: use `lightbulb` icon instead of `ticket`.

src/com/content-event/event-idea.js
```diff
@@ -1,27 +1,26 @@
...
-
-import ButtonBase						from 'com/button-base/base';
-
+import UIButton							from 'com/ui/button/button';
 import $ThemeIdea						from '../../shrub/js/theme/theme_idea';
 
 
 const MAX_IDEAS = 3;
+const canHaveMoreIdeas = (ideas, submitting) => Object.keys(ideas).length + (submitting ? 1 : 0) < MAX_IDEAS;
+const hasSubmittedIdeas = (ideas) => Object.keys(ideas).length > 0;
```
mikekasprzak: I don't have a nice way to suggest this via eslint, but here's my recommendation for how these lines should be formatted.

```js
const canHaveMoreIdeas = (ideas, submitting) => ((Object.keys(ideas).length + (submitting ? 1 : 0)) < MAX_IDEAS);
const hasSubmittedIdeas = (ideas) => (Object.keys(ideas).length > 0);
```

Yes functionally speaking they are the same, but the goal is to try and make the code easier to read with just a glance, by isolating elements with brackets. To keep things sane, the top-most level operation's brackets can be ignored. 

An example:

```js
// this is fine
if ( idea.length > 0 ) {
```
However:
```js
// but this
if ( idea.length > 0 && idea.length <= 64 ) {

// should be
if ( (idea.length > 0) && (idea.length <= 64) ) {
```

The top-level operation in this case is the boolean (`&&`) operations. In the above case, the top-level operation is the lambda (arrow function) operator.

There is an exception.

```js
// this is fine
if ( !this.thirsty && this.health ) {

// this is not
if ( !this.thirsty && this.health > 0 ) {

// it should be this
if ( !this.thirsty && (this.health > 0) ) {
```
So long as only a prefix/suffix operator is used, brackets can be ignored.

src/com/content-event/event-idea.js
```diff
@@ -1,27 +1,26 @@
...
 
 export default class ContentEventIdea extends Component {
 	constructor( props ) {
 		super(props);
 
 		this.state = {
-			"idea": "",
-			"ideas": null
+			'idea': '',
```
mikekasprzak: This should be:
```js
'idea': "",
```

Another style thing I can't convince eslist to do (I got blocked one time trying to make an eslint plugin that enforced some of my _exotic_ style rules. Alas I didn't have time to really get it working or finished).

* `'` should be used if it's something that will never contain a whitespace/newline (i.e. slugs, keys).
* `"` should be used anywhere there is the potential for whitespace or newlines.

src/com/content-event/event-idea.js
```diff
@@ -30,35 +29,34 @@ export default class ContentEventIdea extends Component {
...
+				this.setState({'ideas': r.ideas, 'enableSubmit': canHaveMoreIdeas(r.ideas)});
 			}
 			else {
-				this.setState({"ideas": {}});
+				this.setState({'ideas': {}});
 			}
 		})
 		.catch(err => {
-			this.setState({"error": err});
+			this.setState({'error': 'Error fetching your previous suggestions. Make sure you are logged in.'});
```
mikekasprzak: Should use `"` around `"Error fetching your previous suggestions..."` and not `'`.

src/com/content-event/event-idea.js
```diff
@@ -30,35 +29,34 @@ export default class ContentEventIdea extends Component {
...
-//				if (!e) { var e = window.event; }
-//				if (e.keyCode == 13) { /*e.preventDefault();*/ submitIdeaForm(); }
-//			}, false);
-//		</script>
-
-	textChange( e ) {
-		this.setState({"idea": e.target.value.trim()});
+	textChange( e, isSubmit ) {
+		let idea = e.target.value;
+		if (isSubmit) {
```
mikekasprzak: Prefer:

```js
if ( isSubmit ) {
```

Another style thing I can't convince eslint to enforce. 

The idea is that only function calls don't have spaces between brackets. Function declarations do however. This way you can tell the difference between a function declaration and a call just by glancing.

```js
myFunction( e ) {                 // declaration
        myFunction("something");  // call
}
```

src/com/content-event/event-idea.js
```diff
@@ -30,35 +29,34 @@ export default class ContentEventIdea extends Component {
...
+			idea = idea.trim();
+		}
+		this.setState({'idea': idea, 'error': idea.length > 64 ? 'Suggestion is too long.' : null});
 	}
 
 	onKeyDown( e ) {
 		if (!e) {
```
mikekasprzak:
```js
if ( !e ) {
```

src/com/content-event/event-idea.js
```diff
@@ -30,35 +29,34 @@ export default class ContentEventIdea extends Component {
...
-			var e = window.event;
+			e = window.event;
 		}
-		if (e.keyCode === 13) {
-			this.textChange(e);
+		const isSubmit = e.keyCode === 13;
+		this.textChange(e, isSubmit);
+		if (isSubmit) {
```
mikekasprzak: Prefer:
```js
if ( isSubmit ) {
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
+			this.setState({'error': 'Unexpected error.'});
 		}
 	}
 
-	submitIdeaForm( e ) {
-		var idea = this.state.idea.trim();
-		console.log('submit:', idea);
+	checkDuplicateIdea(idea) {
+		const {processingIdea, ideas} = this.state;
+		if (processingIdea && idea == processingIdea) {
```
mikekasprzak: Prefer:
```js
if ( processingIdea && (idea == processingIdea) ) {
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
+		else {
+			return Object.values(ideas).indexOf(idea) > -1;
+		}
+	}
 
-		if ( idea.length > 0 && idea.length <= 64 ) {
+	submitIdeaForm( e ) {
+		let idea = this.state.idea.trim();
+		//console.log('submit:', idea);
+		if (this.checkDuplicateIdea(idea)) {
```
mikekasprzak:
```js
if ( this.checkDuplicateIdea(idea) ) {
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
+			this.setState({
+				'error': 'Duplicated suggestion! There\'s no point in supplying the same suggestion more than once.'
```
mikekasprzak:
```js
'error': "Duplicated suggestion! The..."
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
+			});
+		}
+		else if ( idea.length > 0 && idea.length <= 64 ) {
```
mikekasprzak:
```js
else if ( (idea.length > 0) && (idea.length <= 64) ) {
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
+			this.setState({'enableSubmit': canHaveMoreIdeas(this.state.ideas, true), 'processingIdea': idea, 'error': null});
 			$ThemeIdea.Add(this.props.node.id, idea)
 			.then(r => {
-				console.log('r', r);
-				this.setState({"ideas": r.ideas, "idea": r.status === 201 ? "" : idea});
+				//console.log('r', r);
+				this.setState({'ideas': r.ideas, 'idea': r.status === 201 ? '' : idea, 'enableSubmit': canHaveMoreIdeas(r.ideas), 'processingIdea': null});
```
mikekasprzak:
```js
this.setState({'ideas': r.ideas, 'idea': (r.status === 201) ? '' : idea, 'enableSubmit': canHaveMoreIdeas(r.ideas), 'processingIdea': null});
```

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
 			})
 			.catch(err => {
-				this.setState({"error": err});
+				this.setState({'error': 'Error processing the request. Make sure you are still logged in.', 'processingIdea': null});
```
mikekasprzak: `"`

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
 			});
 		}
 		else {
-			this.setState({"error": "Problem with length"});
+			this.setState({'error': 'Suggestion is too ' + (idea.length == 0 ? 'short.' : 'long.')});
```
mikekasprzak:
```js
this.setState({'error': "Suggestion is too " + ((idea.length == 0) ? "short." : "long.")});
```
In this case, `"` for consistency, as they are used to build a larger string containing whitespace.

src/com/content-event/event-idea.js
```diff
@@ -67,48 +65,64 @@ export default class ContentEventIdea extends Component {
...
 		}
 	}
 
 	renderIdea( id ) {
-		var idea = this.state.ideas[id];
+		const idea = this.state.ideas[id];
```
mikekasprzak: :+1: I just want to acknowledge that this is good. I started doing this myself (adding consts) thanks to seeing it so often in your code reviews.

src/com/content-event/event-idea.js
```diff
@@ -117,21 +131,51 @@ export default class ContentEventIdea extends Component {
 		return Object.keys(this.state.ideas).map(this.renderIdea);
 	}
 
-	render( {node, user/*, path, extra*/ }, {idea, ideas, error} ) {
+	render( {node, user/*, path, extra*/}, {idea, ideas, error, enableSubmit} ) {
```
mikekasprzak:
```js
render( props, state ) {
    var {node, user/*, path, extra*/} = props;
    var {idea, ideas, error, enableSubmit} = state;
```

src/com/content-event/event-idea.js
```diff
@@ -117,21 +131,51 @@ export default class ContentEventIdea extends Component {
...
 		if ( node.slug && ideas ) {
 			if ( user && user['id'] ) {
-				return (
-					<div class="-body">
-						<h3>Theme Suggestion Round</h3>
+				const ShowError = error ? <div class="content-base content-post idea-error">{error}</div> : null;
+				let ShowSubmit = null;
+				if (enableSubmit) {
```
mikekasprzak:
```js
if ( enableSubmit ) {
```

src/com/content-event/event-idea.js
```diff
@@ -117,21 +131,51 @@ export default class ContentEventIdea extends Component {
...
+							<UIButton onclick={this.submitIdeaForm}>
+								<SVGIcon>suggestion</SVGIcon> Submit
+							</UIButton>
 						</div>
-						<div class="foot-note small">You have <strong>{MAX_IDEAS - Object.keys(ideas).length}</strong> suggestion(s) left</div>
-						<h3>My Suggestions</h3>
+					);
+				}
+				let ShowMySuggestions = null;
+				if (hasSubmittedIdeas(ideas)) {
```
mikekasprzak:
```js
if ( hasSubmittedIdeas(ideas) ) {
```

src/com/content-event/event-idea.js
```diff
@@ -117,21 +131,51 @@ export default class ContentEventIdea extends Component {
...
+					ShowMySuggestions = (
 						<div class="idea-mylist">
 							{ this.renderIdeas() }
 						</div>
+					);
+				}
+				let ShowRemaining = null;
+				const remaining = Math.max(0, MAX_IDEAS - Object.keys(ideas).length);
+				if (remaining > 1) {
```
mikekasprzak:
```js
if ( remaining > 1 ) {
```
(and the lines below this)

src/com/content-event/event-idea.less
```diff
@@ -1,36 +1,78 @@
...
-		& > .-text {
-			padding:0 0.25em;
+			& > .-text {
+				padding:0 0.25em;
+				font-style: italic;
+			}
 		}
 	}
-}
+}
```
mikekasprzak: newline

mikekasprzak:
coding style suggestions aside, this is good. I like that it's like the rest of the site (that was the intent, but I was in a hurry to port over the old code, and never found the time to go back and fix this).

src/com/content-event/event-idea.js
local-minimum: I think there might be a lint rule for not mixing operators that could be used for these... perhaps.

mikekasprzak: Sadly it's for mixing `&&` and `||` (and it's already enabled).

src/com/content-event/event-idea.js
local-minimum: This one will be impossible to keep track of ;)

mikekasprzak: It's more about raising errors for certain `'` uses.

To be fair, it might be more work than it's worth. Sticking to `"just double quotes"` might be lesser of evils.

PR #1570, opened 2018-01-21
[WIP] Theme selection in the meta

local-minimum:
**This builds upon #1569**

This introduces a shrub-tool for the CLI to change theme-related event settings. These settings are stored in the node-meta table as `public` with `b=0`.

Currently the only used setting is the `theme-ideas-limit` which changes the number of ideas that can be suggested by a user.

## Important

This changes the number of default allowed suggestions from 3 to 0. To me this makes much more sense given that there now is an easy tool to set the value of suggestions to anything. It also makes more sense because it allows simple understanding of 0 or lack of value to be meaning that the event doesn't include theme suggestion round. (Can still have selection round).

I aim to extend the shrub-tool with all different theme-related settings in a similar manner to the two that are in there. I aim to use `theme-has-selection` or similar key to denote that the event includes any type of theme-process. Understanding that if it doesn't such tab will not show on the UI and if the event has a theme, it will have been set by the event creator.

In the shrub-tool, I'm planning to make a `ld-default` master setting that does all the needed settings to make theme-suggestion work like previously.

I'm thinking of using date-time stamps rather than theme-stages since the latter can be inferred from the former but not the other way around. That also means we can show schedules for the theme-parts of the event. 

This of course will break compatibility with previous LD on the site, but only in the sense that they would not show any information about the theme-selection in their current state (which should be fine since they don't do that anyhow). It should also be possible to remedy this by using the above `ld-default` in the shrub tool on the existing ld-events.

### DIFF
--- a/public-api/vx/theme.php
+++ b/public-api/vx/theme.php
@@ -342,7 +342,7 @@ switch ( $action ) {
 				if ( $event = validateEvent($event_id) ) {
 					// Is Event Accepting Suggestions ?
 					if ( isset($event['meta']) && isset($event['meta']['theme-mode']) && intval($event['meta']['theme-mode']) === 1 ) {
-						$theme_limit = isset($event['meta']['theme-idea-limit']) ? intval($event['meta']['theme-idea-limit']) : 3;
+						$theme_limit = isset($event['meta']['theme-idea-limit']) ? intval($event['meta']['theme-idea-limit']) : 0;
 						
 						$RESPONSE['response'] = themeIdea_Add($idea, $event_id, $user, $theme_limit);
 						
--- a/src/com/content-event/event-idea.js
+++ b/src/com/content-event/event-idea.js
@@ -121,9 +121,11 @@ export default class ContentEventIdea extends Component {
 
 		return (
 			<div class="-item">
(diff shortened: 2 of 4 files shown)


mikekasprzak:
> This changes the number of default allowed suggestions from 3 to 0. To me this makes much more sense given that there now is an easy tool to set the value of suggestions to anything

This is fair. For compatibility, an entry for this metadata will have to be manually inserted in the data of the events we have run. Not a big deal, but something that needs to be remembered once this code is merged.

> I'm thinking of using date-time stamps rather than theme-stages

Actually states are preferred, since it simplifies other code (`if state > 1 do thing`). It also means things can have 3 states (before/not started, active, after/finished). Automation (another feature that doesn't exist yet) will handle changing states once a calendar date rolls around. State changes are going to be one of the most common automation types (i.e. when `timestamp > value` do `key=value`).

local-minimum:
> This is fair. For compatibility, an entry for this metadata will have to be manually inserted in the data of the events we have run. Not a big deal, but something that needs to be remembered once this code is deployed.

While it probably is good, it shouldn't be needed since the run events don't allow any new theme suggestions anyhow and my updates to the UI makes it so that even if we decide to show what people did suggest after theme rounds, the UI would not say that the had used more suggestions than they had the right to.

PR #1573, opened 2018-01-28
Theme idea previous events & using theme ideas max from the meta

local-minimum:
**Builds upon #1569 and implements what was previously #1570**
* UI uses the event-meta fully to count events.
* Adds event-meta to simulator so that events from the simulator can have themes suggested again.
* Lets user see last 6 event themes from same event group to help them not suggest things that just recently was the theme. This means it's not specific to LD but can be used by any event-group.
* Handles events that didn't have theme.

image

A note on previous LD. They don't need updating for this PR to work with them. There's nothing in the UI that would allow users to see the theme-idea page of an event that has completed, but it is probably good to use the new event shrub tool to set their idea-max to 3 in case of future changes to the UI.


mikekasprzak:
TODO: set `theme-idea-limit` for all existing events.

src/shrub/tools/theme/theme-selection.php
```diff
@@ -0,0 +1,110 @@
...
+	if ($ACTION) {
+		print " ** ERROR: Action '".$ACTION."' unknown.\n";
+	}
+	print "\nACTIONS:\n";
+	print "\t* 'has-selection': If event includes theme suggestions round.\n";
+	print "\t* 'idea-limit': The number of idea suggestions per user.\n";
+}
+
+print "\n";
+?>
```
mikekasprzak: Don't close PHP files that are "all code".

Also file should end with a newline.

local-minimum:
What is blocking this merge?

mikekasprzak:
Database backed up. Entries of `theme-idea-limit` set to `3` for nodes 10, 9405, 32802, and 49883 added.

mikekasprzak:
Sorry for the late merge. I mistakenly thought this did more.

Issue #1381, opened 2017-11-20
Add hotkeys to theme slaughter


zwrawr:
Mention on the site here https://ldjam.com/events/ludum-dare/40/$50300/script-to-add-keyboard-controls-to-theme-slaughter

Also in the same vain as #12

local-minimum:
I think we need to pull in something like https://github.com/NekR/preact-delegate to have hotkeys work. If @mikekasprzak is OK with that, I think it would be a quick fix.

mikekasprzak:
Well it doesn't actually need it, but that is an interesting approach.

Issue #1059, opened 2017-07-24
Theme Slaughter Toggle


pfefferminzz:
![ld-toggle](https://user-images.githubusercontent.com/5459440/28512101-b79ce238-7050-11e7-8eb9-4c6984765771.png)

Please add toggle buttons to make it possible to change/correct a recent theme slaughter decision.

gonutz:
This is the same as [issue 325](https://github.com/ludumdare/ludumdare/issues/325).

pfefferminzz:
Oops. My bad. I searched for "slaughter", but the search did not return anything. :(

gonutz:
It's fine, you can just close this issue and maybe comment on the other one. Issue 325 is rather old but I guess it will go up in priority if enough people comment on it.

Issue #343, opened 2016-12-02
Theme Slaughter: Re-evaluate AccessKeys


mikekasprzak:
See #342. 

Problems:
* Need to use ALT+Key to activate (?)
* Inconsistent with Arrow Key behavior
* QWERTY specific (AZERTY and DVORAK ignored)

PR #1575, opened 2018-01-31
Theme slaughter update

local-minimum:
* Unified style with rest of site.
* Hotkeys for yes and no (not flag because people shouldn't abuse it) #1381, #343 ? 
* Colors on buttons, recent vote history and chart match.
* You can change your mind for a vote by clicking on it. #1059, #325 
* CSS animations to show what is updating and happening.
* Code cleaning
* Pie-chart supporting custom color definitions.

I haven't seen #332 so maybe close that one too.

ld-theme-slaughter2

### DIFF
--- a/src/com/content-event/event-slaughter.less
+++ b/src/com/content-event/event-slaughter.less
@@ -112,6 +112,11 @@
 				width: 57%;
 			}
 
+			& > .-info {
+				font-size: 70%;
+				font-style: italic;
+			}
+
 			@media (max-width: 768px) {
 				width: 100%!important;
 				display: block;
--- a/src/com/content-event/event-slaughter.less
+++ b/src/com/content-event/event-slaughter.less
@@ -30,9 +30,8 @@
 		margin:0.5em 0;
 		vertical-align: middle;
 
-		& button {
-			background: #AAA;
-			color: #DDD;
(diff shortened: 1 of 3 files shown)


src/com/content-event/event-slaughter.less
```diff
@@ -41,22 +45,42 @@
...
-			border-radius: 2.0em;
+			// border-radius: 2.0em;
 			font-weight: 700;
 			outline: none;
+			color: @COL_L;
 
-			&.-green {
-				background:#4C4;
-				color:#8F8;
+			&.-yes{
```
mikekasprzak: space ` {`

mikekasprzak:
I'm not seeing where the zoom effect is being triggered. Let me know what's causing it. We're going to have to tweak it to avoid excessive repaints.

local-minimum:
That's strange. Just tested on FF and Chrome and they work fine. But lets take that in a chat. The load on the client should be fine. I'm running it on a not too cool notebook while also having the server running on it and I don't notice any issues.

mikekasprzak:
This is good, but for reference #332 is still there. To reproduce you need to vote on some themes and pay close attention to the order, then refresh. The order breaks.

PR #1576, opened 2018-01-31
Test stickler

local-minimum:
Adds stickler linting to js, yaml and php.
Could also lint css, but then we might need to change css-linter.

(Ready to be reviewed now)


mikekasprzak:
How do I test this?

PR #1579, opened 2018-02-04
Theme fusion prototype

local-minimum:
ld-theme-fusion-prototype

This prototype doesn't post anything to any api, it is just for evaluating a way to do theme fusion (UI) and a way to find similar themes to ask input from the user.

* Current version randomly picks a theme and finds similar themes based on longest common substring.
* It filters out low scores and picks maximum of 7 similar and presents the group of up to 8 to the user.
* It asks for either 'FUSE' or 'NO FUSE' (non of the themes should be fused with each other).
* It sometimes will present only on option (the random selected theme), this would mean no similar at all in the whole set of themes, and in the future this would trigger it to pick next theme to find similarity group.
* Since it just records the state in the component, refresh page to try again.

### DIFF
--- a/src/com/content-event/event-fusion.js
+++ b/src/com/content-event/event-fusion.js
@@ -2,9 +2,9 @@ import {h, Component}					from 'preact/preact';
 import SVGIcon 							from 'com/svg-icon/icon';
 import NavLink 							from 'com/nav-link/link';
 
-import ButtonBase						from 'com/button-base/base';
+import UIButton							from 'com/ui/button/button';
 
-import $ThemeIdea						from '../../shrub/js/theme/theme_idea';
+import $ThemeIdeaVote					from '../../shrub/js/theme/theme_idea_vote';
 
 
 export default class ContentEventFusion extends Component {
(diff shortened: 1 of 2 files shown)


mikekasprzak:
Oh wow, that's very different than what I was proposing. Interesting.

My interpretation of Theme Fusion is much simpler. Present the user with 2 options:

**[BAGELS] [DONUTS]**

You pick which one you think is the better theme. Alternatively, you can pick the 3rd option, which is to note that they are the same, or flag them.

**[These are the same]** [Flag Bagels an inappropriate] [Flag Donuts as inappropriate]

Flagging though is only necessary if fusion replaces the slaughter. If it comes _after_ the slaughter, and the master list is a subset of the total themes (i.e. 1000 of them), the actual inappropriate themes will be filtered out already.

The key thing to understand is anything we present to a user needs to be as simple as possible. I worry that the above is too complicated. The interface for it is good, but the idea of presenting folks with the responsibility of saying two or more things are the same is potentially quite confusing.

local-minimum:
That's fine. And I could easily swap so that it only asks these questions,  because I agree that this is by complex and not so fun. However I don't see how you would find the themes that are similar just by chance. I thought we wanted help with that.

My main interest with this PR is to see what similarities the scoring algorithm will pick up on true LD themes.

Issue #1636, opened 2018-04-24
"Play my game" spam as comments


mikekasprzak:
Like #1202, but it's potentially worse since it subverts the natural voting.

mikekasprzak:
Added this as a `support bottleneck` since I'm the goto person when folks find this.

local-minimum:
It could look something like this:

![image](https://user-images.githubusercontent.com/8041100/39331867-fd352b1e-49a5-11e8-947b-9e9db84322e0.png)

`onChange` event detects that I've entered a link to my game and it replaces the publish etc buttons with a helpful message.

local-minimum:
It could look something like this:

![ld_comment_nolinks](https://user-images.githubusercontent.com/8041100/39331939-40835df0-49a6-11e8-8d08-eb72e4d6c4aa.png)

The component reacts to and `onChange` event that sees that I've entered a link to my own page and replaces the publish etc. buttons with a helpful info box.

mikekasprzak:
Neat. Does it stay there or is dismissable/go away at any time.r?

SecretPocketCat:
Couldn't there be a button to mark comments as unhelpful? Clicking it would be the opposite of the 'Love btn' and would lower the commenters coolness. 
(Does the love button raise it? I'm actually not entirely sure what it does).

Issue #1625, opened 2018-04-23
@tag auto name suggestion breaks users with spaces


jasonleekennedy:
for example

if someone types @ jason - it suggests @ jason kennedy, but but the user name per the system is @ jason-kennedy.  Appears to prevent the tag from getting detected for notifications and the tag links to the wrong or non-existent user

mikekasprzak:
I've been getting a bunch of these.

https://twitter.com/HarryLightGames/status/988504843122806791

JackWesleyNelson:
Yeah, I noticed this as well. I thought someone mistagged another, but when I went to correct them I realized that while it still autocompletes the name it doesn't actually associate a correct link. (And only the first break of the name is hyperlinked)

PR #1639, opened 2018-04-27
Restricts some comment behaviour + bonus

local-minimum:
ld_comment_nolinks
(Post above in gif is by the same author as the one writing in the gif)

Gives feedback / hides submit buttons if:

1. Duplicating last posted comment (valid for all pages)
2. Posting url to your most recent authored post (this could probably be made more exact in pointing to most recent game for all makers on the game)

Exceptions to 2.:
* This only is a limit on games
* You are allowed to post whatever if you are one of the authors of the post.

And as a bonus, it might solve comment at-name autocomplete for those with more complex names.

Deals with issues #1636 #1625

### DIFF
--- a/src/com/content-comments/comment-autocomplete.js
+++ b/src/com/content-comments/comment-autocomplete.js
@@ -93,7 +93,7 @@ class Autocompletions extends Component {
 			return;
 		}
 		let {selected} = this.state;
-		selected = this.getOptions(selected ? selected : match).sort((a, b) => b.score - a.score)[0].name;
+		selected = this.getOptions(selected ? selected : match).sort((a, b) => b.score - a.score)[0].slug;
 		this.executeSelect(selected);
 	}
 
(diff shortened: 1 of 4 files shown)


zwrawr:
I against stopping people from linking to there game.
The vast majority of people that link to their game are not being malicious, it seems unfair to stop them because a very small number of people are spamming.

Here's an example of a legitimate use https://ldjam.com/events/ludum-dare/41/ludum-dare-41-placeholder-1/post-your-games-for-me-to-rate , and there are many of these threads.

There is going to be backlash if you stop people linking to there games.

local-minimum:
Link stopping was restricted to only be on other people's games. Not on posts for exactly that reason. @zwrawr do you still think that's bad then?

mikekasprzak:
Can we get the auto complete fix isolated to it's own PR? I got yet another report of the autocomplete problem today.

local-minimum:
@mikekasprzak I think those changes are isolated to https://github.com/ludumdare/ludumdare/pull/1639/files#diff-f2d8e3005d0a2d9ba2c416c60bf7b0e9 but you should build and test it locally. I didn't have any usernames that would match the problem, but the fix seemed reasonable.

PR #1648, opened 2018-05-08
Danger Mode!!

zwrawr:
As per usual ( we should really get round to automating this) the last week is normally in danger mode. Where games that are just shy of being ranked are prioritized.

Currently we have just under 50% of games ranked.
image

Looking good for getting the same ranked % as the last two events
image


mikekasprzak:
Rather than have to change it, I wouldn't mind making the "smart" filter smart enough to include game near the cut-off just to push them over. It is called "smart" after all. ;)

PR #1668, opened 2018-05-28
Uses slug not name for autocompletes

### LINKED ISSUE #1625: @tag auto name suggestion breaks users with spaces

local-minimum:
To make sure you can correctly autocomplete those with a space, autocomplete now uses slugs rather than names for what is inserted into the text.

Tested by making a user with a space and using it.

Resolves #1625

### DIFF
--- a/src/com/content-comments/comment-autocomplete.js
+++ b/src/com/content-comments/comment-autocomplete.js
@@ -93,7 +93,7 @@ class Autocompletions extends Component {
 			return;
 		}
 		let {selected} = this.state;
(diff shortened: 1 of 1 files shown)


mikekasprzak:
Great. Out of curiosity, do all the fields need to be made in to the slug?

local-minimum:
I just changed those that deal with the pasting. It still suggest actual name in the widget

Issue #1626, opened 2018-04-23
Impose restrictions on submission


mikekasprzak:
Now that the website is stable enough for submissions, we should impose restrictions on submissions:

* Time - Only allow Compo to be selected until 2 hours after compo deadline, or if published before then (for all those people that click the wrong event and contact me after)
* Team Size - if >1, disallow the Compo

local-minimum:
I briefly researched this and while it isn't hard, the structure of the code makes it very impractical since getting the authors list is part of `<ContentSimple />`  while the logic behind what is allowed need to go in the `<ContentItem />`.

Which again makes me wish we used the Container pattern https://medium.com/@learnreact/container-components-c0e67432e005

And https://redux.js.org/ (which does have a selling-point that it is tiny 😉 )

Reworking the structure of the code base would be a huge undertaking, but I think it would make it more accessible to contributions if organized in a standard and very predictable way and I think we would work much faster adding features to it too.

mikekasprzak:
A change I've been making to a bunch of code that greatly simplifies things is taking code that looks like this.

```html
<ContentSimple node={node} user={user} featured={featured} path={path} />
```

And changing it in to this.

```html
<ContentSimple {...props} />
```

local-minimum:
Can we close this now?

PR #1670, opened 2018-05-30
Limit options for Compo submission

local-minimum:
This removes the `Compo` option from items if there's more than one author. It partially deals with #1626 but remains:

- [ ] Remove `Compo` option after Compo has ended.
- [ ] Hide ability to add or remove people to game while Compo is selected.
- [ ] For some reason, though the cache is supposed to be invalidated, authors don't update on the game without a refresh of the page. Is this a bug we should add to the list? At least if it was left in edit-mode to add a friend to the team.

image

### DIFF
--- a/src/com/content-item/item.js
+++ b/src/com/content-item/item.js
@@ -358,10 +358,20 @@ export default class ContentItem extends Component {
 		);
 	}
 
+	getNumberOfAuthors() {
+		const {node} = this.props;
+		if ( !node.meta || !node.meta['author'] ) return 1;
+		if (node.meta['author'].filter(n => n == node.author).length == 0) {
+			return node.meta['author'].length + 1;
+		}
+		return node.meta['author'].length;
+	}
+
 	render( props, state ) {
 		props = Object.assign({}, props);
 		let {node, user, path, extra, featured} = props;
 		let {parent} = state;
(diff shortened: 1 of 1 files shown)


mikekasprzak:
I went ahead and made a change (#1672). Counting the number of authors is now a global function `node_CountAuthors` inside `/internal/lib/node.js`. Simple node information extractions should go there instead of being implemented every time they're used.

> Remove Compo option after Compo has ended.

This should be its own issue. There's a few points I need to clarify about this issue specifically, and if done improperly it actually creates more work for me (i.e. users e-mailing me that they accidentally submitted in the wrong event).

> Hide ability to add or remove people to game while Compo is selected.

Also its own issue. Related, the key thing to takeaway is we shouldn't be hiding, but rather it should be made clear this is not allowed. i.e. graying out the button to add people. Also the current method of adding/removing users from a team will likely be removed once the real way is added (i.e. doing it right from the Edit page), so it's probably not worthwhile fixing what's there now, but rather finishing the proper way.

> For some reason, though the cache is supposed to be invalidated, authors don't update on the game without a refresh of the page. Is this a bug we should add to the list? At least if it was left in edit-mode to add a friend to the team.

Yes this is a bug. One or more of the caches that need to be invalidated are not being invalidated. This can also be its own issue.

Issue #1450, opened 2017-12-04
Use `&nbsp;`  or non-line breaking in general for team-member names.


local-minimum:
![image](https://user-images.githubusercontent.com/8041100/33577749-50b6c892-d944-11e7-86dd-e56cb35c822e.png)

Breaking after the parenthesis looks quite wonky.

A small issue though.

sgstair:
Related to this, we should look into adding the concept of "at-mention-possessive" e.g. "@sgstair's" as it's used in notifications, at the moment there's unnecessary space between before the 's

povrazor:
- [ ] Wrap names in author line with a span featuring whatever that `no-wrap` tag is.
- [ ] Remove gap from around `@name`'s, so `'s` will work as expected

PR #1671, opened 2018-05-30
Author names no break

local-minimum:
image

Was more than enough to put `Super` on top line but now name is kept intact. (Just cropped heavily to make sure it worked).

#1450 Initial issue resolved. Nothing done to space around @ - names because that doesn't seem to relate to how they are used here.


mikekasprzak:
This seems to include the contents of #1670 too.

Issue #1657, opened 2018-05-20
Make "Anonymous Comments" a checkbox instead of a button


mikekasprzak:
This way we can include a "help blob" that tells them things like #1262

local-minimum:
I'll work on this next week after #1678 is merged (so I can make a common `<InputCheckbox />` component.)

PR #1680, opened 2018-06-02
Using checkboxes for item edit

local-minimum:
image

# Related Issues

* #1657 Is resolved by this

### DIFF
--- a/src/com/content-item/item.js
+++ b/src/com/content-item/item.js
@@ -15,6 +15,8 @@ import ContentCommonNavButton			from 'com/content-common/common-nav-button';
 
 import InputStar						from 'com/input-star/star';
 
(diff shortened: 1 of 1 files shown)


mikekasprzak:
Ha! This is not exactly what #1657 is about (the issue is that posting anonymously is a separate publish button, and that clutters space), but hey, this seems a decent change too. :+1:

mikekasprzak:
Looks like the shown state of the out-out checkboxes might be backwards. Before pulling your patch, I opted out of "Complexity", and after the upgrade my state became this:

screenshot from 2018-06-06 13-28-24

local-minimum:
I'll check it I missed something.

local-minimum:
As commit message said, there was some complexity in how the opt out data was prepared I didn't understand why, so I removed it and now it works.

PR #1693, opened 2018-06-12
Rate Games in feed

### LINKED ISSUE #1644: Lead users landing on the frontpage into the game rating process
The rating process should be given more prominence on the front page.

Right now it's a little bit hidden, and a couple of clicks deep depending on the flow you follow.

This makes the news feed a great target for people trying to advertise their game as it will have much more presence when posted here, which encourages people to spam the news feed with posts- which does seem to be an effective strategy if you look at the number of ratings received/given for these games.

If the front page lead people into the rating process and actively advertised games which needed rating the power of these posts would be reduced, and would also reinforce an important aspect of ludum dare- which is the rating of games.

This is an example mock up of how the front page flow could be adjusted:

image

Here the top 4 entries from the smart list would be inserted into the top of the news feed as an auto-generated blog post.

This gives people a brief reminder of how the rating process works, and offers up several games which are currently most in need of rating.

The number of games you still need to rate this jam is also given as a prompt to the user that they should continue to rate games.

This post could also appear as the top item each time you click 'more' at the bottom of the news feed, with a new set of games randomly picked from the top 100 smart list to continue to try to attract the users attention into the rating process.

local-minimum:
ld_rategames

# Features

* Adds an item into the root page feed on top, before all other things, that will display 3 games at a time.
* Only visible when there's a featured event on which there is currently voting.
* It randomly selects 3 out of the top 30 on the smart feed. If it wasn't random there would be a big risk of it being useless since many times after a while the top 1-3 games are unplayable to the majority of participants due to requirements or other issues.
* It collapses and expands like other posts in the feed.
* There's a refresh that re-picks the random 3 out of the same 30 games to be shown (without fetching). Re-picking happens up to 7 times. After that it does a new fetch (you see the spinner breifly in the gif) and you have 7 new iterations of 3 random picked games.
* Adds a button/link to the main page for browsing games to vote on.

# To discuss

* I'm not very happy with selection of `SVGIcon`s for reshuffle and all games in the lower right. Which is also why I wrote the text next to them.
* In my box, for some reason, the event won't list all games (not if I go there the normal way either), so I'm not 100% the `href` is correctly built from the `featured.slug`.
* There's no text about the importance of playing and rating, but I'd prefer to keep the noise down.
* If there's news, it will come below this however important the news is.
* There's no separate default for small screens such as mobile

# Related issues

* Resolves #1644

### DIFF
--- a/src/com/content-timeline/timeline-rategames.js
+++ b/src/com/content-timeline/timeline-rategames.js
@@ -37,13 +37,17 @@ export default class TimelineRateGames extends Component {
     if (props.featured != nextprops.featured) this.getGames(nextprops);
   }
 
+  event_canRate(featured) {
+    if (!featured || featured.type !== 'event') return false;
+    if (!featured.published || !featured.meta) return false;
+    return featured.meta['can-grade'] == "1";
+  }
 
   getGames(props) {
     console.log('getGames', props);
-    const {featured} = props;
-    if (!featured) return;
+    if (!this.event_canRate(props.featured)) return;
 
-    const {id} = featured;
+    const {id} = props.featured;
     const methods = ['parent', 'superparent'];
     const types = ['item'];
     const subtypes = ['game'];
@@ -87,7 +91,7 @@ export default class TimelineRateGames extends Component {
   }
 
   render(props, {expanded, feed, pick, error, loading}) {
-    if (!props.featured) return null;
+    if (!this.event_canRate(props.featured)) return null;
     const HeaderClass = cN('content-common-header');
     const MainClass = cN('content-base', 'content-common', 'rate-games', !expanded && 'minimized');
 
--- a/src/com/content-timeline/timeline-rategames.js
+++ b/src/com/content-timeline/timeline-rategames.js
@@ -16,15 +16,30 @@ export const randomPick = (maxExclusive, n) => {
   return ret;
 };
 
+const N_RESHUFFLE_BEFORE_RELOAD = 6;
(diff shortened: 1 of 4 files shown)


zwrawr:
Yea that looks like back icon not a refresh icon. Even just the [Spinner](https://github.com/ludumdare/ludumdare/blob/master/src/icons/icomoon/spinner.svg) might be better.

Looking on the icomoon website, they do have refresh/reload icons that we could use.

local-minimum:
Was considering the spinner, maybe that would be better yes. Easy to swap if Mike says that's better.

Yngwarr:
It's worth mention that these should be 3 games user haven't rate yet. And also, why to limit it with top-30? Wouldn't it be more useful to make a choice completely random or at least top-100 random? It seems that games in the top of smart rating getting too much attention.

local-minimum:
30 is just an arbitrary number. There's a limit to how many you can get from the API and I think it's not worth the added complexity to go beyond that. I leave it to Mike to tweak that number.

As for not including already rated, that's part of another issue (to not do that in general) so it will resolve when someone fixes the backend for the smart filter.

As for smart getting too much attention, I don't quite understand the point. The purpose of smart and design of ldjam is to promote and reward good community behavior. In this case, playing, rating and writing good feedback. When a game gets rewarded by people playing back those who made a serious effort, it drops down in smart rank. The exact weighting is ofc up for discussion. I think there are several issues relating to smart scores waiting for someone to tackle them.

Issue #1422, opened 2017-12-02
Allow joining more than one game


winniehell:
Especially for people making assets (music / graphics) or who want to take part in compo and jam with a different game, it would be nice to be able to join multiple games.

winniehell:
Correction: It is possible to join multiple games. It is just not possible to start a new game when you are already part of one.

local-minimum:
I'm guessing that what happen when you click the my game button and all other things like towards what game your votes count need to be taken into account too.

winniehell:
Yes, "My Game" only links to one of them and "New (blog post)" only allows creating blog posts for one of them.

winniehell:
For Ludum Dare 40, I was part of two games in case that helps debugging.

mikekasprzak:
Yeah, multiple game support isn't there yet. Behind the scenes it works,
but the user experience is vague and unclear.

How it's working right now: My Game and New Post points to the last game
from the active session that was edited. If you refresh, it may change
(caches pending). Long term this isn't practical, but it helped us work
around an issue for folks that joined before realizing they should be part
of a team.

How it should work: that depends.

For most users, the current assumption mode is fine. In general we do still
need fine-toothed control over which game we're posting to (outside event
weekends).

For a hand full of others though (sound designers, floaters at jam sites,
etc), they almost need a way to say "hey I'm doing multiple games".

We could always display all games from the current event you're associated
with, except when it's just one. That just means the next request will be
"deleting the bad game", for those users with 2 games accidentally.
Ultimately this is fine. But because of potential orphaned
posts/history/metadata/links/references, it needs to be handled with care.

PR #1695, opened 2018-06-18
Game team building

### LINKED ISSUE #1448: Improve team building (was Team Questions)
Howdy;

So I have found my way to https://ldjam.com/users/josefnpat/following

And I want to make sure the team is set up right. How do I make sure this is the team for the current jam and not my previous jam? Or is it the same list, so to speak?

Also, I can't seem to "Remove From" previous members. It just has the js popup, and then does nothing, even if I refresh the page.

local-minimum:
ld_buildteam

# Features
* Select what friends you want to add and remove from a simple drop-down.
* Removes old slightly buggy way of adding friends to game.
* Works directly and live and integrates with the preview of the item.
* Only affects items, not posts.

# To discuss

Right now all authors get access to the team-building feature. But it's a question of who should have this right. For now you can remove yourself if you are not main author and you can remove everyone but yourself if you are main author. Everyone gets to add from their own friends (does the api even allow that? Should it allow that?)

# Related issues
* Resolves #1448 
* #1422 may be fixed by this or are there other limitations?

### DIFF
--- a/src/com/content-item/item-teambuilding.js
+++ b/src/com/content-item/item-teambuilding.js
@@ -1,0 +1,139 @@
+import {h, Component} from 'preact/preact';
+import ContentCommonBody from 'com/content-common/common-body';
+import ContentLoading	from 'com/content-loading/loading';
+import ContentError	from 'com/content-error/error';
+import UIIcon	from 'com/ui/icon/icon';
+import UIButton	from 'com/ui/button/button';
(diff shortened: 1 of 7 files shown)


mikekasprzak:
Fantastic! Fixed a couple bugs related to having no friends (or a half-completed friendship) and some wording, but overall an excellent much-needed fix. :+1: :+1:

Issue #1699, opened 2018-07-31
Typing a "y" in the password field when logging in closes dialog and produces an error message


Nicate:
The following error message is shown:

Session Expired
Oops!!
It looks like your session expired (and yes, it's probably a bug).
Log in again to continue what you were doing.

I gots them Firefox 61.0.1 on them Windows 7. Dunno what else to tell ya. Didn't happen during previous Ludum Dares.

Nicate:
Oh yeah, this also means I can't log in. If this happens for other people it's possible a lot of people can't log in.

Samusoidal:
What page are you logging in from? If it's the Theme Slaughter page, try logging in from the homepage as it doesn't use keyboard shortcuts Y and N.

Nicate:
Whew, at least that means I can join in two weeks :D

local-minimum:
Same as #1700 then?

Samusoidal:
Yup, duplicate.

local-minimum:
I'm the guilty one, but with the organization of the site's code I think the most tangible way is to on that page continuously check if user is still checked in an disable Y/N key capture when user gets logged out.

Unfortunately Mike is MIA due to pollen and I have quite a few PR up so I'll at least wait until those have been merged.

mikekasprzak:
#1592

Issue #1592, opened 2018-04-12
Theme Slaughter hotkeys listening while not logged in/a dialog is open


mikekasprzak:
Ah shoot. It looks when you're visiting the Theme Slaughter page, the `Y` and `N` hotkeys try to do stuff, even while you're not logged in.

https://twitter.com/SuciuTiberius/status/984472862168965120

At the very least, we need to stop this with a check, 'cause it's triggering that error message.

Eventually this is going to have to be switched to something focus related. Once the Search popup is added, you're going to want to be able to type `Y` and `N` keys.

local-minimum:
What specifically is the issue here? The minimal solution so to say?

Is it the error message in itself, that it should say that the user needs to log in?
I'm thinking the client doesn't really know if the user is logged in unless it tries to vote on a theme.

PR #1718, opened 2018-08-13
Unbinds hotkeys when logout detected

### LINKED ISSUE #1700: Can't login if username/email contains the letter 'n'
My username/email contains the letter 'n' which causes a dialouge to pop up when pressed.

**Text:**
> **Oops!!**
> It looks like your session expired (and yes, it's probably a bug).
> Log in again to continue what you were doing.

**Example:**
https://gfycat.com/CorruptExcellentAngwantibo

local-minimum:
It's not the most beautiful fix, basically checking if `$Node.GetMy` is allowed once per minute. And if not say logged out. Also if a vote gets 401 it sets logged out. And when setting logged out it removes the event listener.

Resolves #1700 #1699 #1592

### DIFF
--- a/src/com/content-event/event-slaughter.js
+++ b/src/com/content-event/event-slaughter.js
@@ -6,6 +6,7 @@ import SVGIcon 							from 'com/svg-icon/icon';
 import UIButton							from 'com/ui/button/button';
 
 import $ThemeIdeaVote					from 'shrub/js/theme/theme_idea_vote';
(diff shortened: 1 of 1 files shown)


local-minimum:
I realized I could do this without adding any calls, just checking the status on the attempted vote and unbind key capture event if voting gets a non 200. So should be safe to merge/won't add any new load on the site.

PR #1734, opened 2018-08-15
Attempt at fixing nan-stars

### LINKED ISSUE #1717: NaN next to rating
image

I'm not sure what it's supposed to be, but I'm not a fan. Suggest hiding that part of the component until whatever it is can be calculated.

local-minimum:
Resolves #1717 (hopefully, hard to check as things are now)


ajayyy:
When it shows NaN, the value not always actually zero, sometimes there has already been a rating.

PR #1736, opened 2018-08-15
Batching of User and Featured Event data (WIP)

mikekasprzak:
By request, here is the current working version of the latest batching improvements.

**This code is unfinished**. It breaks compatibility with the old `featured` data (i.e. `what_node`, `what`, etc).

### DIFF
--- a/public-api/vx/node2.php
+++ b/public-api/vx/node2.php
@@ -218,7 +218,8 @@ api_Exec([
 	// if not logged in, where will be blank
 	$RESPONSE['where'] = nodeComplete_GetWhereIdCanCreate(userAuth_GetID());
 }],
-// Figure out what's featured, and what you made for it
+// Figure out what event is featured, and if you are athenticated, what you made for it.
+// NOTE: This is slightly different than the original behavior, but it combines several requests in to a single batch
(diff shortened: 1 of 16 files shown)


public-api/vx/node2.php
```diff
@@ -218,17 +218,56 @@ function nodeAPI_Filter( $nodes, &$more_nodes ) {
 	// if not logged in, where will be blank
 	$RESPONSE['where'] = nodeComplete_GetWhereIdCanCreate(userAuth_GetID());
 }],
-["node2/what", API_GET | API_AUTH | API_CHARGE, function(&$RESPONSE, $HEAD_REQUEST) {
-	$parent_id = intval(json_ArgShift());
-	if ( !$parent_id )
+// Figure out what's featured, and what you made for it
+["node2/what", API_GET | API_CHARGE, function(&$RESPONSE, $HEAD_REQUEST) {
```
local-minimum: why is there no longer auth required to figure out 'what you made'?

local-minimum: alright I understand why it works but with that comment on the line above it looks a bit scary..

mikekasprzak: Fair enough, the description can be improved.

PR #1741, opened 2018-08-17
Theme idea help no dupe

local-minimum:
This builds upon/includes #1740 

It introduces a live "Please consider if these other themes aren't similar enough to what you are writing that you don't have to suggest it yourself".

It's not perfect but it's a start:

ld_simliarideas

It's hard to capture what may be similar ideas for themes without understanding them so I weighted two common metrics into a single score. It has issues on very short suggestions (2-3 characters long), saying that everything is very similar to everything else short. That might be fine. It will still find if there's an identical or near identical match and present it first.

The other issue is that the `<input onchange/>` function seems a bit buggy or at least sub-optimal. It lags one character in firing the events so if you had previously typed `hell` and just entered an `o`, you would see `hello` but the event to fire would be `hell`. If you defocus the input or add another character like ` `. It would fire `hello`. This means that the suggestions are lagging behind all the time (can sort of be seen in the gif).

It uses historical data of LD-event, recent events (on new site), everyone's suggestions thus far on this event.

If you suggestion according to the metrics is awfully close to an existing, they are colored red. If what is similar is an old theme, it will say `LD 32` or such (as you see towards the end of the gif).

### DIFF
--- a/public-api/vx/theme.php
+++ b/public-api/vx/theme.php
@@ -24,7 +24,7 @@ function GetEventNodes() {
 
 function validateEvent( $event_id, $optional = false ) {
 	global $RESPONSE;
-	
+
 	if ( $event_id !== 0 ) {
 		/// Check if $event_id is on the master list of event nodes.
 		if ( !in_array($event_id, GetEventNodes()) ) {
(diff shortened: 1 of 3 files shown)


ajayyy:
```
Indirect interaction
Guardian
Construction/destruction (sheep)
Preparation – Set it up, let it go
Infection
Random
Light and darkness
Growth
Swarms
Moon/anti-text
Build the level you play
Chain reaction
Weird/unexpected/surprise
Minimalist
The tower
Roads
Advancing wall of doom
Caverns
Exploration
Islands
Enemies as weapons
Discovery
It’s dangerous to go alone! Take this!
Escape
Alone (kitten challenge)
Tiny world
Evolution
You are the villain (goat)
Minimalism (potato)
10 seconds
You only get one
Beneath the surface
Connected Worlds
Entire Game on One Screen
An Unconventional Weapon
You are the Monster
Growing/two button controls
Shapeshift
Ancient Technology
One Room
```

Plain text list of old themes if you would like.

local-minimum:
Thanks, I think `$ThemeHistory.Get()` gives the same but there might be a point in storing it directly in the HTML rather than doing the extra API request. Mike can decide on that.

mikekasprzak:
Yeah http://api.ldjam.com/vx/theme/history/get was written specifically for
this. Speaking of, if you can think of anything data-wise that should be
added or tweaked (I'm thinking we should have a field for ID, so history
can make reference to locally run events) let me know. We definitely need a
shell script for adding to the list. I may tweak the internals a bit to
better parallel the true data also (i.e. include a path or slug).

ajayyy:
Oh ok, I didn't know there was an API to get old themes.

local-minimum:
@mikekasprzak I parse the ld-number now because I wanted it short (as you can see in the gif), if endpoint had it that would be sweet.

PR #1745, opened 2018-08-20
Showing icons on jam in games feed

### LINKED ISSUE #1480: Include something that lets you tell if a game was made solo or as part of a team
food

Now that the Jam has taken over (nearly 2:1), it would be nice to know if a Jam game was made solo or as part of a team from glance.

Originally that was going to mean showing avatars of all contributors, but instead we could use the Person and People SVG iconography in a corner.

local-minimum:
User = one author
Users = not one author

image

Resolves #1480


mikekasprzak:
Awesome! I wasn't 100% sure it was going to work, but yeah it actually looks really good!

PR #1747, opened 2018-08-20
Dealing with your strange request response

local-minimum:


local-minimum:
The comments-fix is about timing when switching user and trying to load comments before `user` has been gotten.

PR #1753, opened 2018-08-27
My Grades in main event nav

local-minimum:
Putting `My Grades` button in the main event nav when mode reaches the opening of voting.

image

The me-nav lingers only when actually visiting the my grades-page otherwise it is removed. It's rather useless at the moment so perhaps even better dropping it entierly?


local-minimum:

PR #1758, opened 2018-08-29
Change email fields to type="email"

### LINKED ISSUE #1380: Forgot password input type should be `email`
Input field should use `email` type to display appropriate keyboard on mobile.

paulloz:
Fixes #1380


local-minimum:
FYI everything still goes through Mike, me approving is just a signal to Mike that I think it is sound

PR #1759, opened 2018-08-29
Events/group page

local-minimum:
image

Features:
* Shows any node text with markdown
* If you are not in `/events` then it shows a `<- To parent`.
* It sorts the stuff in the group so that events and groups are listed (groups before events) at the bottom of the page
* Pages (such as "Rules" and "Hosting Guide" etc) become buttons below the body of the node.
* If you are in the events-group that has the name `Ludum Dare` then the theme-history from previous site will be added to the listing too. It will link to the old site's results for that event (but since there were no results in the early events they don't get links).
* If you are logged in, it shows the games you made for each event (supports multiple games per event!)
* For events it shows theme (if exists) and the status of the event.
* For groups it shows the first sentence of that groups node.


zwrawr:
Probably need to add an entry in here https://github.com/ludumdare/ludumdare/blob/master/src/public-ludumdare.com/main-ld.js#L575

local-minimum:
One issue was that I added stuff the wrong way in the DB
Might be the only issue really, but still have some kinks to work out.

local-minimum:
This works now btw, if that wasn't clear. But as with many of the PR, I guess it is nicer to wait until after the event and things settle down a bit.

local-minimum:

Issue #1698, opened 2018-07-24
Make the Rules/Events more accessible


Samusoidal:
It's kind of hidden there on the event page, and I've noticed that new jammers seem to have a hard time finding/understanding the rules. 

There should also be a 'tutorials' page where new users can see how to open issues, contribute to the project, and use the site (creating teams, etc).

ajayyy:
I agree, there should be a link in the header or sidebar.

local-minimum:
@paulloz is looking into this one

paulloz:
Would a **Guides** section in the sidebar be good enough?  
We can link the **rules** and the **game hosting guidelines** from there, and the **tutorials** @Samusoidal is talking about when they're written.  

![firefox_2018-08-29_18-57-48](https://user-images.githubusercontent.com/437025/44803043-90203a80-abbd-11e8-86b3-94561fde09d2.png)

local-minimum:
The one concern is that `https://ldjam.com/events/ludum-dare/rules` are the rules for LD-jams only, and the dream is to one day support the miniLD again. But well if you ask me this is much of an improvement and if we get to supporting more than main LD then we can deal with it later.

Samusoidal:
I like the idea of a Guides section in the sidebar but it needs to stand out like a sore thumb in a bucket of sardines.
Maybe you could put a fill behind it? Rounded edges, light gray from the palette, etc. That would be eye catching and denote that it's important.

local-minimum:
I'm quite happy with the style in the image. We are planning other ways to point people to the rules too like the new rules questions when you submit and https://github.com/ludumdare/ludumdare/issues/1752

local-minimum:
But we like `<SVGIcon>` so perhaps add one that fits for the rules. You find them in http://192.168.48.48/dev_icons.php or something close to that

PR #1760, opened 2018-08-30
Sidebar guides

paulloz:
You can add the links you want in the sidebar and/or change their icons, as discussed in #1698.

firefox_2018-08-30_17-05-56

### DIFF
--- a/src/com/sidebar/guides/guides.js
+++ b/src/com/sidebar/guides/guides.js
@@ -1,0 +1,30 @@
+import {h, Component}	from 'preact/preact';
+import SVGIcon			from 'com/svg-icon/icon';
+
+export default class SidebarGuides extends Component {
+	constructor() {
+	}
+
+	render(props, state) {
+		var Links = {
+			"Rules": {"icon": "article", "href": "/events/ludum-dare/rules"},
+			"Game Hosting Guide": {"href": "/events/ludum-dare/hosting-guide"}
+		};
+
+		return (
+			<div class="sidebar-base sidebar-shortlist sidebar-guides">
+				<div class="-title _font2">
+					<SVGIcon baseline>question</SVGIcon> <span class="-text">Guides</span>
+				</div>
+				{Object.keys(Links).map(key => {
+					return (
+						<a class="-item" href={Links[key].href}>
+							{Links[key].icon != null ? <SVGIcon baseline gap>{Links[key].icon}</SVGIcon> : null}
+							<strong>{key}</strong>
+						</a>
+					);
+				})}
+			</div>
+		);
+	}
+}
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ node_modules/
 /src/shrub/docs
 /sandbox/ld_event.*
 package-lock.json
+*.code-workspace
+.vscode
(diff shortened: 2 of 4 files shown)


local-minimum:
Where did the Twitch integration end up?

paulloz:
Oh it is still [just below](https://github.com/paulloz/ludumdare/blob/sidebar-guides/src/com/view/sidebar/sidebar.js#L70), obviously it can be put above or anywhere.

src/com/sidebar/guides/guides.js
```diff
@@ -0,0 +1,31 @@
...
+			"Game Hosting Guide": {"href": "/events/ludum-dare/hosting-guide"}
+		};
+
+		return (
+			<div class="sidebar-base sidebar-shortlist sidebar-guides">
+				<div class="-title _font2">
+					<SVGIcon baseline>question</SVGIcon> <span class="-text">Guides</span>
+				</div>
+				{Object.keys(Links).map(key => {
+					const icon = Links[key].icon != null ? <SVGIcon baseline>{Links[key].icon}</SVGIcon> : null;
```
local-minimum: If you add `gap` to the SVGIcon tag you can skip the space injection further down

local-minimum:
I think it's good below. I just wondered since we discussed its position

paulloz:
Well, I don't really mind the placement. What's bothering me is that it always looks like it's being part of the section above it.  

On another note, I updated the `gap` thingy.

src/com/sidebar/guides/guides.js
```diff
@@ -0,0 +1,30 @@
+import {h, Component}	from 'preact/preact';
+import SVGIcon			from 'com/svg-icon/icon';
+
+export default class SidebarGuides extends Component {
+	constructor() {
```
zwrawr: This constructor should probably have `super(props);`
 in or not be there. Although I'm pretty sure it doesn't make an actual difference.

src/com/sidebar/guides/guides.js
```diff
@@ -0,0 +1,30 @@
...
+		};
+
+		return (
+			<div class="sidebar-base sidebar-shortlist sidebar-guides">
+				<div class="-title _font2">
+					<SVGIcon baseline>question</SVGIcon> <span class="-text">Guides</span>
+				</div>
+				{Object.keys(Links).map(key => {
+					return (
+						<a class="-item" href={Links[key].href}>
```
zwrawr: This should probably be a `NavLink` rather than an `a` tag, right?

src/com/sidebar/guides/guides.js
```diff
@@ -0,0 +1,30 @@
...
+							{Links[key].icon != null ? <SVGIcon baseline gap>{Links[key].icon}</SVGIcon> : null}
```
zwrawr: Would it make sense to set a default icon (maybe `article` or `file-text`) rather than no icon. If they all had icons they'd all be aligned.

zwrawr:
This works and looks good. Should make these guides much easier to find!

paulloz:
While I'm happy to see activity around here, I must admit that after that much of a review time I probably don't have a usable dev environment for this project anymore.  
I could certainly setup everything once again. Although I guess it'd even be quicker for someone else to make those changes because I don't know when I'll be able to re-download vagrant images etc.

zwrawr:
That's fine, #1819 has your changes

Issue #1752, opened 2018-08-26
Making joining event process more informative and intuitive


local-minimum:
Clicking the join-button should in the background create the placeholder game-post as it does now, but it should __not__ redirect you to that page since you most of the time won't have business there yet instead it should redirect you to an educational page on the event that explains the parts of the event, what they are for and what is expected of the user in each part.

When the user presses `Publish` on their game they should probably be redirected back to the same page to the heading of how the voting works.

Something along these lines for LD:

# Welcome to Ludum Dare `{number}`!

If this is your first time here. Please take your time reading through these instructions. It can be quite overwhelming the first time and it is very easy to miss important pieces. At least read this first section now and maybe come back later as the event progresses.

Ludum Dare is a game jam were thousands of people around the world make games either as teams or solo. It is a process that is spit into three main parts:

* __Theme selection__ which is optional but really good for hype. Together we come up with ideas, prune them and vote on which will become the theme. It is kind of customary that a large fraction of the participants will find the theme annoying, but that's part of the challenge.
* __Game making__ starts when Mike announces what is the theme for the event. You will have either 48 or 72 hours to make your game. Different rules apply, so check them before you you start.
* __Playing others' games, rating them and giving them good feedback__ is as important and mandatory as the making of games. This is a three weeks period during which we check out what others' made and help them grow and become better developers. _We expect __everyone__ to play and rate at least 20 games._

## I'm in post

A lot of people write a post when they join. Like you just did! They say that they will be partaking and usually list what tools and techniques they plan to use. Maybe they include a screenshot of where they'll be working, the pet that will be helping them... and so on.

If they plan to do the jam and have some own tools they're planning to use, this is a good time to share them since the rules stipulate that pre-made tools need to be publicly available. 

These posts can look a bit boring, but they also serve the purpose of orienting everyone in new tools that could be interesting to try out. Maybe you should try a new game engine?

## Theme selection

There are some general guiding principles about what is a good theme:

* It should not be hurtful or derogatory.
* It should be inspiring for everyone. This is a tricky thing since it means that the theme needs to work for those that want to work in VR, AR, mobile, 3D, 2D, text only, and audio only at the same time. In other words, most themes that restrict or describe mechanics are highly problematic.
* It should also be fun to play 20 games, preferably a lot more than 20. If the theme only have one very narrow interpretation that makes the vast majority of all submissions play very similarly, the voting period becomes quite tedious.
[...]

PR #1765, opened 2018-09-03
Join redirects to welcome (based on event node meta)

local-minimum:
If the you click join event and then the create game button and the event your game/item is created for  has a meta `event-join-page`, then instead of redirect to edit of your game you get redirected to that page. So my suggestion is copying the text of #1752 into a page named `Welcome` having `ludum-dare`-group as parent. Then make a meta between next LD event and that page.


local-minimum:

PR #1750, opened 2018-08-23
Login fixing

local-minimum:
Resolves #266 #1501 

ld_login

* Can now save login+password pairs in Firefox, Chrome and probably Safari.
* The autofill works in Firefox and Chrome when using mouse and/or keyboard to select stored login
* Can now submit form from both input-fields, not just the password field.

PR #1766, opened 2018-09-03
Just the login fix

local-minimum:
As #1750 but without the extra files...


local-minimum:
@mikekasprzak if we get back to working on the UI maybe we can take care of this and my other old PR?

Issue #1614, opened 2018-04-20
Reserved Usersnames are X'd out


mikekasprzak:
When registering a reserved username, for some reason the reserved usernames are X'd out. That said they still work.

mikekasprzak:
Okay, this is a bit of a weird one. Feel free to pass on it if you're not a fan.

Long story short, I got a report that when you login and you are on our reserved list (a table we populated with a list of email+username combos), even if the data is correct, the little helper icon (an X or checkmark) shows as an X.

This issue isn't so much about fixing (yet), it's about reproducing the problem and understanding if it's simply a misinterpretation (i.e. insert some dummy "reserved" users in to the database, and try to use their name with a non-matching e-mail address). I haven't personally confirmed that this is an issue. It could have been browser-specific, or what I'm thinking is it was due to focus (i.e. the icon doesn't change until you change focus away from the field containing the username). It's very possible the user that reported it simply misunderstood that the little :heavy_check_mark: icon would only pop up after focus changed. The reason for this was because it meant the servers wouldn't get hammered with every keystroke.

So again, the task right now is just to determine if the way it works right now "works", and if it was simply a misunderstanding of what "it works" means. Again I seem to remember designing it in such a way that I only hit the server with a request when the field focus was lost.

PR #1767, opened 2018-09-03
Create user fix

local-minimum:
Fixing issues discussed in #1614 

Now the X and the checkmark work as expected. Giving error/hints what is wrong too. Also fixed code-style


local-minimum:

Issue #1762, opened 2018-09-01
Add `alt`,  `title` and `aria` on all buttons.


local-minimum:
We have major accessibility issues on our site. As soon as a button collapses (like the navs and such do when changing the width of the browser) you kind of have to guess what they do.

PR #1769, opened 2018-09-04
Removing dead code, adding title to nav links.

local-minimum:
Cleaning up some unused code that was seriously confusing that it still existed.
Otherwise main objective is to give simple tooltips for the links because the way the site displays on a narrow view now the nav buttons are really not helpful just being icons without any text. If you agree with this PR then I can continue with the other buttons that behave similarly.


local-minimum:
This does some of the #1762

local-minimum:

PR #1780, opened 2018-11-29
Moscow Standard Time != Mountain Standard Time

nurumaik:
I think you should add global time below calculated local, because I'm sure it does still have a lot of bugs


zwrawr:
fyi on hover, a more generic time is shown. But that doesn't help too much if you don't know about it.

Issue #1821, opened 2019-06-02
Site broken on IE11


zwrawr:
The site is currently broken on IE 11. 

![Screenshot from 2019-06-02 16-08-01](https://user-images.githubusercontent.com/18352787/58763292-bb6ef600-8550-11e9-87f1-abc429a75ba8.png)

```
HTML1300: Navigation occurred.
ldjam.com
[constructor]
[fetchData]
[fetchUser]
[fetchRoot]
[fetchFeatured]
[fetchUser] caller_id: 0
[fetchUser] Done: 0
[fetchNode]
[fetchFeatured] + 139254
Potentially unhandled rejection [1] TypeError: Unable to get property 'type' of undefined or null reference
   at e.prototype.matchRoute (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:106158)
   at prototype.getCurrentRoute (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:105485)
   at e.prototype.componentWillReceiveProps (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:106922)
   at setComponentProps (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:96894)
   at buildComponentFromVNode (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:98919)
   at idiff (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:94192)
   at innerDiffNode (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:95429)
   at idiff (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:94609)
   at diff (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:93692)
   at renderComponent (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:98169)
[fetchFeatured] My Game(s): 
   "[fetchFeatured] My Game(s):"
   [
      length: 0
   ]

[fetchFeatured] - 139254
Potentially unhandled rejection [2] TypeError: Unable to get property 'type' of undefined or null reference
   at e.prototype.matchRoute (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:106158)
   at prototype.getCurrentRoute (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:105485)
   at e (http://ldjam.com/-/all.min.js?v=4497-6a7d992:1:105249)
[...]

PR #1822, opened 2019-06-02
Copy short link button

### LINKED ISSUE #1788: Implement a way to use ldj.am
Similar to this awesome firefox extension: https://addons.mozilla.org/en-US/firefox/addon/ld-shortlinkener/

zwrawr:
Closes #1788 

## Features
Adds button at the end of post titles. This only shows when on the post's page and not on the main feed.   
image

When clicked this copies the correct short link to clipboard. If successful the icon is replaced with a  :heavy_check_mark: for two seconds. If it fails the icon changes to :negative_squared_cross_mark: .
Screenshot from 2019-06-02 15-34-52

Copy to clipboard functionality from https://gist.github.com/lgarron/d1dee380f4ed9d825ca7 we might want to use https://github.com/lgarron/clipboard-polyfill if this simple solution is too buggy but so far it works well.

## Issues
- [x] ~~Not sure how to get the correct shortener link form js, it only appears to be defined in PHP, but not in the section where PHP constants are copied to js. See the below links for those files.~~
  
 https://github.com/ludumdare/ludumdare/blob/e7c1dc6b8d155e774d454374898bfb8a7574203c/src/public-url.redirect/main-redirect.php#L15
    https://github.com/ludumdare/ludumdare/blob/e7c1dc6b8d155e774d454374898bfb8a7574203c/src/main/main.php#L64

~~I'm not sure how to add a constant for the correct link shortener (ldj.am vs url.ludumdare.org). [this section of js](https://github.com/ludumdare/ludumdare/compare/master...zwrawr:short-link-button?expand=1#diff-d93231edb9fe5368a7151a230625e2f9R58) needs updating with the correct shortener URL.~~
  
------------------------------------
- [ ] Not tested on ie (see #1821) or mobile browsers (works on chrome, firefox, edge41 and opera)

### DIFF
--- a/src/com/button-clipboard/clipboard.js
+++ b/src/com/button-clipboard/clipboard.js
@@ -1,0 +1,74 @@
+import {h, Component}					from 'preact/preact';
+import ButtonBase						from 'com/button-base/base';
+import SVGIcon							from 'com/svg-icon/icon';
+
+export default class CopyToClipboardButton extends ButtonBase {
+
+	constructor( props ) {
+		super(props);
+		this.onClick = this.onClick.bind(this);
+
+		this.state = {
+			"data": props.data,
+			"state": "IDEL"
+		};
+	}
+
+
+
+	onClick(e) {
+
+		let info = e.target.getElementsByTagName('div')[0];
+
+		this.copy(this.state.data).then( () => {
+			console.log("Copied to clipboard");
+
+			if (this.state.state === "IDEL") {
+				this.setState({"state": "SUCCESS"});
+				window.setTimeout(() => {
+					if (this.state.state === "SUCCESS") {
+						this.setState({"state": "IDEL"});
+					}
+				}, 2000);
+			}
+
+		}).catch((e) => {
+			console.log("Error copying short link to clipborad", e);
+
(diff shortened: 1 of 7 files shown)


local-minimum:
Does the first state with the chain image say copy to clipboard on hover? Otherwise it might be a bit magical and vague what is happening. In general I think it would be best with a temporary tooltip that says "Short link copied to clipboard" and a negative message for failing.

zwrawr:
Whoops, I totally forgot about messages because I was messing with having a card slide into the bottom. Before I decided that should be its own thing.
infocard

I'll get on adding them.

zwrawr:
I've added tooltips with different messages depending on the state and a way to override the "IDEL" tooltip.

mikekasprzak:
> Not sure how to get the correct shortener link form js, it only appears to be defined in PHP, but not in the section where PHP constants are copied to js. See the below links for those files.

I would do basically the same thing seen in `ludumdare/src/main/main.php` for the static domains. It has a little lookup table of domains with matching URL shorteners, then a couple lines later does a lookup.

zwrawr:
I've done what you suggested and it seems to work.

PR #1836, opened 2019-10-07
skew towards feedback

mikekasprzak:


local-minimum:
As in promote good feedback in who gets sorted  to the top on the smart filter?

local-minimum:
Ah now I, too tried after the ludum to realize it was a merge and not an issue lol

Issue #1837, opened 2019-10-09
Build is failing


zr3:
Travis CI is showing the build as failing, and I ran into the same issues that it's reporting when following the dairybox instructions to try to set this up locally to take a crack at a few issues, so that I could be the third person in a row to commit to this repo who has "zak" somewhere in their name.

It is caused by wildcard node dependencies that brought in breaking changes to parts of the make process (e.g. stricter LESS linting rules), which is why a bunch of stuff seemed to randomly fail after a few months of no commits/builds.

PR #1838, opened 2019-10-09
Fixed the build

zr3:
Updated packages, fixed some breaking changes, added some linter rules for ubiquitous less issues and fixed some easy issues.

This addresses issue #1837


package.json
```diff
@@ -4,25 +4,25 @@
...
-    "rollup-plugin-node-resolve": "*",
-    "rollup-plugin-nodent": "*",
-    "uglify-js": "*",
-    "preprocess-cli-tool": "*",
-    "less": "^2.0.0",
-    "less-plugin-autoprefix": "*",
-    "stylelint": "^8.2.0",
-    "stylelint-config-recommended": "^1.0.0",
-    "svgo": "*"
+    "babel-eslint": "^10.0.3",
```
zr3: These are all the current latest versions

src/config/eslint.config.json
```diff
@@ -1,4 +1,5 @@
 {
+	"parser": "babel-eslint",
```
zr3: https://stackoverflow.com/questions/36001552/eslint-parsing-error-unexpected-token is the issue that this addresses

mikekasprzak:
Okay, I'll need a block of time to test these changes on the live server. 99% sure I'll have to update the packages. I'll try to do it in the next couple days.

zr3:
Sounds good to me -- and yeah, I agree that it would definitely be good to merge this when there's not a lot of traffic to the site, if possible. FWIW I was able to spot-check the home page and the main tabs to verify that it was working and that the js and css compiled correctly, but I haven't set up the database yet, so I didn't go any deeper than that.

I also didn't look at any of the sites that aren't ldjam.com/ludumdare.org if they're relevant, still working on familiarizing myself with the project 😊

local-minimum:
@zr3 there's a helper script that will simulate an event which should set up the databases.

But also @mikekasprzak, isn't deployment manual? In that case couldn't we merge it into master without causing any issues and at the same time making it easier for me people to poke around the code?

nicolasbize:
Just checked out the code for the first time and ran into the build issue straight away.
Setting a correct parser for ESLint did the trick indeed.

I'm curious, how do you guys do deployments / prod db changes?

PR #1839, opened 2019-10-12
Fix #109: add notification indicator in favicon

### LINKED ISSUE #109: Suggestion: Show number of alerts in FavIcon (like Gmail)
I'm pretty sure no browser supports SVG favicons, so these need to be generated. GENERATE THEM! DO NOT MAKE THEM BY HAND!

Generate about 30 icons, and start doing "30+", "40+", above that to about 100.

Ideally, a simple PHP script that given a file and a string outputs a PNG (?). Then run the script several times.

nicolasbize:
This is a small fix for #109 that doesn't rely on generating favicons:
- stays in sync with the # displayed in the bubble throughout time and user actions.
- displays the original favicon when there are no notifications
- displays 1-9 according to the # of unread notifications
- displays "9+" for more than 9 notifications.

Tested in Chrome, Safari and Firefox:

chrome-2
chrome-9plus
firefox-3
click

### DIFF
--- a/src/com/view/bar/bar.js
+++ b/src/com/view/bar/bar.js
@@ -155,28 +155,28 @@ export default class ViewBar extends Component {
 		// remove existing icon if any
 		const el = document.querySelector("link[rel='shortcut icon']");
 		if (el) el.parentNode.removeChild(el);
-		// reload LD favicon and paint it
-		const canvas = document.createElement('canvas');
-		canvas.setAttribute('width', 64);
-		canvas.setAttribute('height', 64);
-		const context = canvas.getContext('2d');
 		const icon = new Image();
 		icon.setAttribute('src', '/favicon.ico');
 		icon.onload = () => {
-			// LD favicon as background
+			// reload LD favicon and paint it
+			const canvas = document.createElement('canvas');
+			canvas.setAttribute('width', icon.naturalWidth);
+			canvas.setAttribute('height', icon.naturalHeight);
+			const context = canvas.getContext('2d');
+			// place LD image as background
 			context.drawImage(icon, 0, 0);
-			// We only draw text for # > 0
+			// draw text when # notifs > 0
 			if (notificationCount > 0) {
 				const text = notificationCount > 9 ? "9+" : String(notificationCount);
-				// Dark backdrop for contrast
-				context.fillStyle = '#0d4263';
+				// fill with a dark backdrop for contrast
+				context.fillStyle = '#6f7984';
 				context.rect(text.length > 1 ? 14 : 38, 30, 50, 50);
 				context.fill();
-				context.fillStyle = '#fff';
+				context.fillStyle = '#eef2f7';
 				context.font = 'bold 36px tahoma';
 				context.fillText(text, text.length > 1 ? 16 : 40, 64);
 			}
-			// Replace favicon
+			// replace favicon
 			const link = document.createElement('link');
 			link.type = 'image/x-icon';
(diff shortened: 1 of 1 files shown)


src/com/view/bar/bar.js
```diff
@@ -145,6 +146,45 @@ export default class ViewBar extends Component {
...
+		const icon = new Image();
+		icon.setAttribute('src', '/favicon.ico');
+		icon.onload = () => {
+			// LD favicon as background
+			context.drawImage(icon, 0, 0);
+			// We only draw text for # > 0
+			if (notificationCount > 0) {
+				const text = notificationCount > 9 ? "9+" : String(notificationCount);
+				// Dark backdrop for contrast
+				context.fillStyle = '#0d4263';
```
local-minimum: The colors should probably be from the color-theme (when running the dev environment you can find them here http://ludumdare.org/dev/palette)

src/com/view/bar/bar.js
```diff
@@ -145,6 +146,45 @@ export default class ViewBar extends Component {
...
+	 * Shows individual numbers until 9, then 9+ for more.
+	 */
+	refreshFavIconNotificationCount() {
+		const notificationCount = this.state.notifications || 0;
+		// remove existing icon if any
+		const el = document.querySelector("link[rel='shortcut icon']");
+		if (el) el.parentNode.removeChild(el);
+		// reload LD favicon and paint it
+		const canvas = document.createElement('canvas');
+		canvas.setAttribute('width', 64);
```
local-minimum: Currently the favicon is 64x64 but it could well be that this is open to change and by that time it could become a bit obscure why it starts to look strange, especially if we would deliver icons based on the device requesting it. So if it is possible it would be better to get the width and height from the `el` above.

local-minimum: or even better from the `icon` below.

src/com/view/bar/bar.js
nicolasbize: Great idea! Fixed in 71d9966

src/com/view/bar/bar.js
nicolasbize: Using icon's size in 71d9966.

However note that the positioning / text size is still hard-coded and will look broken/weird the day that you decide to change the favicon to another size. I tried to play around with different icon sizes (16x16, 32x32 all the way to 256x256) to have something fully generic but it seems like the tradeoff between code readability and genericity wasn't quite worth it here. Unless you foresee that the favicon will change several times in the future, I'd advise to only tackle this specific problem when it presents itself to avoid over-engineering.

However, this is my first time contributing in this codebase so I'll adapt to whatever philosophy you guys have adopted so far.

Thanks for the review!

local-minimum: I agree that it's better to keep the code simple, and now the only not looking strange will be the placement and size of the number and that will at least point directly to this feature needing an update.

It's a bit of a pity that the color hex codes need to be duplicerad here, but I don't see any way way too get around that. So the code looks good to me, but it's Mike who finally decides on all

nicolasbize: I actually originally thought to retrieve those values from the DOM, aka
`window.getComputedStyle(document.querySelector(".view-bar")).backgroundColor`
which works fine but becomes brittle as it adds a hard dependency on the dom structure.

Open to include that change for both colors if you feel that it's better.

local-minimum: I think it is better like it is now. Such change would be brittle and quite magic.

PR #1840, opened 2019-10-12
Fix #1081: Remember that you clicked the X on the "noob" popup

### LINKED ISSUE #1081: Remember that you clicked the X on the "noob" popup
The NOOB popup is large and annoying even to me. We should store a key in the localstorage that keeps it hidden even when logged out.

nicolasbize:
This adds a new variable `isNoobHeaderDismissed` in the local storage to remember the fact that a guest has dismissed the noob panel in the Feed page.

bug-1081

### DIFF
--- a/src/com/page/root/root.js
+++ b/src/com/page/root/root.js
@@ -25,14 +25,32 @@ import PageDevPalette 					from 'com/page/dev/palette';
 
 import HeaderNoob						from 'com/header/noob/noob';
 
+const STORAGE_HEADER_DISMISSED = 'isNoobHeaderDismissed';
+
 export default class PageRoot extends Component {
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			isNoobHeaderDisplayed: (!props.user || !props.user.id) &&
+				!localStorage.getItem(STORAGE_HEADER_DISMISSED),
+		};
+	}
+
+	dismissNoobHeader() {
+		localStorage.setItem(STORAGE_HEADER_DISMISSED, 'true');
+		this.setState({
+			isNoobHeaderDisplayed: false,
+		});
+	}
+
 	render( props ) {
 		const {node, user, path, extra, featured} = props;
 		let Dummy = <div />;
 
 		let ShowIntro = null;
-		if (!user || !user.id) {
-			ShowIntro = <HeaderNoob featured={props.featured} />;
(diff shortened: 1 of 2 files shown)


src/com/page/root/root.js
```diff
@@ -25,14 +25,32 @@ import PageDevPalette 					from 'com/page/dev/palette';
...
 import HeaderNoob						from 'com/header/noob/noob';
 
+const STORAGE_HEADER_DISMISSED = 'isNoobHeaderDismissed';
+
 export default class PageRoot extends Component {
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			isNoobHeaderDisplayed: (!props.user || !props.user.id) &&
```
local-minimum: Do we still need these first two checks? If we have it in local storage then why bother with the logic about the user being logged in or not?

nicolasbize: If we remove the first two checks then the popup will be visible for the users that are logged in until they close it. Logged-in users do not see the popup today so this would bring a bit of regression.

nicolasbize:
That's a good edge case to think about! On the other hand, if you wanted a new different message seen and dismissable by all, would it necessarily replace this noob message or be a new placeholder panel?

I'm thinking that not knowing now what message or what behavior you might need in the future is probably reason enough to not try to write any code to support that usecase until it presents itself.

I can definitely make the changes though if you still feel like it's the right approach, but I'd advise on the cautious side regarding premature engineering.

Thanks for the review, you're always very thorough and thoughtful!

src/com/page/root/root.js
local-minimum: True, though only until they dismiss it, then they'll never see it again. But it really doesn't matter so much, I'm just used to having cleaner render components (using react and redux) at work and I was hoping having the local storage flag be an excuse to remove the references to `user`, which I don't think the noob help text component has any business interacting with. But you are right that the feeling of regression might be too annoying.

local-minimum:
You are also right that should we need to reset the users' settings so we consider everyone noob we could just as well just change the local storage key at that time.

src/com/page/root/root.js
nicolasbize: I definitely think that you've nailed down the philosophy of react/redux using lean components throughout the front-end code and I really like the component breakdown that you've done so far. Let me know if you'd rather go with the regression and I'll remove the user checks.

Cheers

PR #1841, opened 2019-10-13
Fix #1307: Duplicate theme suggestions submission

### LINKED ISSUE #1307: Duplicate theme suggestions submission

nicolasbize:
Add a new check against possible duplicate theme suggestion within existing transaction.
Also ensure that both path (with theme limit and without) go through the transactional path to prevent duplicates on both sides.

This addresses the final comment in #1307

### DIFF
--- a/src/shrub/src/theme/theme_idea.php
+++ b/src/shrub/src/theme/theme_idea.php
@@ -176,6 +176,17 @@ function themeIdea_Count( $event_id, $user_id = 0, $threshold = null, $query_suf
 	}
 }
 
+function themeIdea_CheckExisting( $event_id = 0, $user_id = 0, $theme = "" ) {
+	return db_QueryFetchValue(
+		"SELECT count(id)
+		FROM ".SH_TABLE_PREFIX.SH_TABLE_THEME_IDEA."
+		WHERE node=? AND user=? AND theme=?
+		LIMIT 1",
+		$event_id,
+		$user_id,
+		$theme
+	) > 0;
+}
 
 function themeIdea_CountOriginal( $event_id, $user_id = 0, $threshold = null, $query_suffix = ";" ) {
 	if ( $user_id ) {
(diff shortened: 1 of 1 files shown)


src/shrub/src/theme/theme_idea.php
```diff
@@ -421,11 +424,13 @@ function _themeIdea_AddWithLimit( $idea, $event_id, $user_id, $limit ) {
 		if ( empty($result) ) 
 			throw new Exception();
 
-		$count = themeIdea_Count($event_id, $user_id);
-		if ( $count === false ) 
-			throw new Exception();
-		if ( $count > $limit ) 
-			throw new Exception();
+		if ( isset($limit) ) {
```
local-minimum: I'm not that versed in the PHP code, but what benefit does it have to only throw the exception here in the case of exceeding the limit? On the client side it might look like the theme wasn't registered, but it looks to me like it still would be? Don't we need to remove the inserted row in `$result`?

nicolasbize: First the work that you did in https://github.com/ludumdare/ludumdare/pull/1569 ensures that the user never gets to this point in the first place through the UI (nice job BTW!). The only way you can bypass the code you wrote to get to this place in the PHP code is to forge the API request outside of the standard UI flow.

In this code, every statement here is wrapped inside a single SQL transaction that gets rolled back in case an exception is thrown. So basically if anything goes wrong in any of the flow, the entire thing gets rolled back: nothing gets created and an error is returned. It just ensures a certain level of integrity.

Note that this is still imperfect because outside of the INSERT, we are not locking down any table, so we are actually not sure that 2 count reads might not be done at the same time and then proceed with the INSERTs. I do think that this should be good enough for now though as it provides data integrity against a potential single user doing stuff outside of the UI.

local-minimum: Great, neither PHP nor databases are really my strengths

PR #1927, opened 2021-08-29
Fixed race condition where comments would not load sometimes

### LINKED ISSUE #1912: Comments not loading
Comments often don't load for me. It just sits there with this forever. Refreshing the page often fixes it but it can take a few tries.
Screenshot_20201011_192756
I am signed in so that message is also wrong.

marioBonales:
Resubmitting because I changed the branch, since I want to work on different issues
Fix #1912 

There is an error if the user has not loaded when the comments render for the first time, this is due to ```getComments``` referencing the user and trying to get the Id of the object that is null at the time

The reason why this was not surfaced in the JS console is because is being caught by the catch for the ```$Comment.GetByNode``` call, but when I try to print it inside the catch this is what I get:

```js
///src/com/content-comments/comments.js:102
.catch(err => {
    console.log(err);
    this.setState({'error': err});
});
```
Screenshot from 2021-08-28 18-50-10

My fix is to extract the code that uses the user object to a different function, and execute it only if the user is already loaded, otherwise I added the same call to ```componentWillUpdate``` with a check to see if the user was updated from null

Some minor changes that I also made: 
I hid the 'Sign in to post comments' message if the user is null, since it means the user has not loaded yet (When there is a not logged in user, we actually have a User object, but with Id 0)

I also put a min-height on the comments so that the spinner is not cramped in the headline, this is how it looks:
Screenshot from 2021-08-28 18-43-50

There was also a bracket that was causing a warning when running make

A bit out of scope, so I have no problem reverting these changes if you find them intrusive, so let me know

### DIFF
--- a/src/com/content-comments/comments.js
+++ b/src/com/content-comments/comments.js
@@ -31,6 +31,13 @@ export default class ContentComments extends Component {
 		this.getComments(this.props.node);
 	}
 
+	componentDidUpdate(previousProps) {
+		//If the user was fetched after the component has rendered, the state needs to be updated
+		if (!previousProps.user && this.props.user) {
+			this.updateUser(this.props.user);
+		}
+	}
+
 	genComment() {
 		return {
 			'parent': 0,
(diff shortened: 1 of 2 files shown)


mikekasprzak:
Great, thanks for the commit.

My week is pretty full (plus I need to unbreak some things on the live servers), so it may not be until next week when I merge and push this. I did a quick build with this and it seems good, I just want to finish unbreaking the live servers before I merge it.

marioBonales:
> Great, thanks for the commit.
> 
> My week is pretty full (plus I need to unbreak some things on the live servers), so it may not be until next week when I merge and push this. I did a quick build with this and it seems good, I just want to finish unbreaking the live servers before I merge it.

No worries, I'll be available in case it needs any changes, just let me know

mikekasprzak:
Okay, the fix is now live on [ldjam.com](https://ldjam.com). It looks like it's working. :+1: 

Thanks again for the contribution. :smile:

PR #2098, opened 2022-03-30
Use correct page index for voting round check

winniehell:
fixes https://github.com/JammerCore/JammerCore/issues/1391

### DIFF
--- a/src/com/content-event/event-list.js
+++ b/src/com/content-event/event-list.js
@@ -236,7 +236,7 @@ export default class ContentEventList extends Component {
 			if ( user && user['id'] && lists && votes ) {
 				Body = page ? this.renderList(page) : null;
 			}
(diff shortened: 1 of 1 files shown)


src/com/content-event/event-list.js
```diff
@@ -236,7 +236,7 @@ export default class ContentEventList extends Component {
 			if ( user && user['id'] && lists && votes ) {
 				Body = page ? this.renderList(page) : null;
 			}
-			else if ( lists && !lists['1'] ) {
```
winniehell, line 239: this always checked if the **first** round has started—not the own we are currently viewing

winniehell:
I'm unsure how to setup the project locally, so I have only tested the fix in the browser's debugger on minified code.

mikekasprzak:
No problem. The changed version makes more sense than what's there.

PR #2211, opened 2022-10-15
Sort nodes by published/modified *and* id

pschichtel:
this makes the sorting stable even if entries are modified/published in the same second.

Also add additional composite indices. these might not be worth it. Also the original indices might be obsolete.


pschichtel:
@mikekasprzak anything?

pschichtel:
@mikekasprzak I've rebased this and resolved the conflicts, any chance to get this in now?

PR #2212, opened 2022-10-15
Add option to make node sorting stable

pschichtel:
This is an alternative to #2211.

This makes the sorting stable even if entries are modified/published in the same second. It adds a query parameter to allow sorting by both the published/modified and the id column without changing the default behavior.


pschichtel:
@mikekasprzak anything?

InstantExcel:
I just came here to give this a thumb's up.   I don't know how badly the api gets hammered by people trying to query multiple times and de-dupe the results, this'd save a bit of overhead, with a stable result set.

pschichtel:
@mikekasprzak I've rebased this and resolved the conflicts, any chance to get this in now?

PR #2224, opened 2023-04-01
Anti-spam, part 1

mikekasprzak:
- **DATABASE CHANGES**! Rerun `./db-create.sh`
- Added trust scores to nodes and comments
- check for a positive trust score when adding posts to the homepage feed
- stop users with negative trust scores from changing any content
- added an underscore to database fields that are redundant (derived from other data). i.e. `_superparent`
- fixed an outstanding bug with IP address checks and CDNs. May cause you to log-out when roaming
- now logging IP addresses during account creation and login
- dropped SH_NODE_LINK table (obsolete)
- VERSION_TAG's are now VERSION_DETAIL's, to avoid confusion with tagging
- Magic score's are now called value's


pschichtel:
> now logging IP addresses during account creation and login

depending in your log retention, this might be a GDPR issue, unless you don't store them completely or hash them.

Issue #1748, opened 2018-08-22
Karma system may promote bad feedback behavior


local-minimum:
I saw talk on discord that there are users who give low quality feedback, and still have very good karma. I'm thinking that it could be that even low quality feedback picks up a love here and there and so if you blast it across the page you might gain more than if you write high quality appreciated feedback.

I have an idea on how to deal with this, but I need to investigate the current code

mikekasprzak:
I had someone mail me complaining about users that "liked everything" in a
post, and how this could be used to cheat.

local-minimum:
Simple fix would be

```
raw_feedback = min(42, love_total / comments_count *  10)
```

This would cap karma at 42, it's the general quality of your comments up to the cap of 4.2 hearts per comment.

local-minimum:
So if you are a spammer, chances are you'll stay down about less than ten. Only thing this doesn't do is give you credit for having written many comments. But on the other hand, the other aspects of smart does that already in part because it tracks grades given, and so it would be open to only commenting really good on a few games and earning a whole bunch.

mikekasprzak:
I think eventually my intent with the smart filter was to make karma from
previous participants worth more, as a means of offsetting abuse, as it is
most likely come from first-timer accounts.

Anyway, in general I think this needs more research. I'll try to remember
to post a more useful transcript of the reported issue in the morning.

bloc97:
@mikekasprzak Our team's game was affected by someone abusing the system and I think we sent you that e-mail.

I think the current smart balance system can be abused by griefers to target specific games.
Especially this line: 
`$raw_given_feedback = commentLove_CountBySuperNodeNotAuthorKnownNotAuthor($node['parent'], $node['id'], $authors);`
Lets say a griefing team has 3 people. If that team of were to specifically target another game, they could each leave a feedback then like every other feedback.

That would add 31.5 to the targeted game's `smart_against` balance.
(3 * 20 + 3) / 2.0 = 31.5.

Lets' assume that the targeted team has 20 feedback received and 40 given feedback, their initial smart balance would be:
(sqrt(40 * 100 / 20) * 10) - 100 = 41.42;

Then lets have our greifers come and like every comment and leave 3 dummy comments.
(sqrt(40 * 100 / (20 + 31.5) * 10) - 100 = -72.13;

[...]

PR #2267, opened 2023-09-25
Remove penalty for karma given to comments on a user's game

HoneyPony:
If I understand the code right, this removes any penalty for "hearts given to comments that users other than a game author left on that game's page," without otherwise affecting the score calculation.

This should fix the latter half of the discussion in #1748 .

It does not do much about the originally raised point made by the author of that issue, as unfortunately that issue ended up discussing two separate things.


nyairobi:
Gentle reminder that this is still a problem and the fix is one click away.

